PR #2, opened 2020-02-20
Parse full kernel version string

leodido:


fntlnz:
Thank you!!! LGTM

P.S: we need poiana here

PR #80, opened 2021-01-16
Improve Debian builder error output

maxgio92:
**What type of PR is this?**

/kind feature

> /area pkg

Improve the outputs the Debian builder provides when not finding kernel header deb packages.

The error message that the builder returns when not finding kernel headers is augmented with extra group (now for amd64 arch and common). For instance:

```shell
driverkit docker -c ./debian_5.9.0-1-amd64_1.yaml
INFO using config file                             file=./debian_5.9.0-1-amd64_1.yaml
FATA exiting                                       error="kernel headers not found for extra groups: ([\"amd64\"])"
```


maxgio92:
Maybe we can discard this, with #82 merged @leodido What do you think?

leodido:
IMHO yes

maxgio92:
Agree

PR #87, opened 2021-02-01
Docker network and dns flag support

jfarrell:
Adds NetworkMode and DNS config for docker containers started by driverkit.

**What type of PR is this?**

/kind feature

/area build

/area cmd

Adds new flags to the driverkit docker command to enable docker --network and --dns to be set for the build time container that is run to create the given driver/probe. For systems with docker not in host/bridge mode there is no way for the build container to reach out and curl the needed dependencies, with the addition of --network and --dns flags this config can be passed to the docker run api in order to setup the NetworkMode and enable the container to successful run and build the driver/probe.

Tested manually, when looking into the docker subcommand test setup there are changes needed in order to support error reporting and moving the use of Fatal from within the subcommand to report success/failure differently than it does today. This refactor will result in larger changes to the given root/k8s/docker commands and was excluded and is out of scope in this specific PR and is better suited for a focused PR around testing the subcommands better. 

Adds new flags to the docker subcommand that are optional, will not break existing functionality.

```release-note

```

### DIFF
--- a/cmd/docker_options.go
+++ b/cmd/docker_options.go
@@ -1,0 +1,44 @@
+package cmd
+
+import (
+      "fmt"
+
+      "github.com/creasty/defaults"
+      "github.com/falcosecurity/driverkit/validate"
+      "github.com/go-playground/validator/v10"
+      logger "github.com/sirupsen/logrus"
+)
+
+var dockerOptions *DockerOptions
+
+// DockerOptions represent the configuration flags for the driverkit docker subcommand.
+type DockerOptions struct {
+      DNS          []string `validate:"gte=0,dive,ip" name:"docker dns" default:[]string{}`
+      NetworkMode  string   `validate:"omitempty" name:"docker network"`
+
+      configErrors bool
+}
+
+// NewDockerOptions creates an instance of DockerOptions.
+func NewDockerOptions() *DockerOptions {
+      o := &DockerOptions{}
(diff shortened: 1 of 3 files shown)


cmd/docker_options.go
```diff
@@ -0,0 +1,44 @@
...
+      "github.com/creasty/defaults"
+      "github.com/falcosecurity/driverkit/validate"
+      "github.com/go-playground/validator/v10"
+      logger "github.com/sirupsen/logrus"
+)
+
+var dockerOptions *DockerOptions
+
+// DockerOptions represent the configuration flags for the driverkit docker subcommand.
+type DockerOptions struct {
```
leodido, line 15: Probably the docker options are already available in the go source code of docker.

Could we use that so to inherit the exact same behavior?

leodido (commented):
~~Would you please signoff the commit?~~

Thanks for submitting this (and signing-off your commit) :)

Left a comment

cmd/docker_options.go
leodido: Some places where I'd look for them are:

- https://github.com/docker/cli/tree/master/opts
- https://github.com/moby/moby/tree/master/api/types/container

A quick search brought me this:

https://github.com/docker/cli/blob/88c6089300a82d3373892adf6845a4fed1a4ba8d/cli/command/container/opts.go#L631

PR #107, opened 2021-11-03
new(driverkit): allow configuring the module driver and device name

LucaGuerra:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/area build

/area cmd

/area pkg

> /area docs

> /area tests

The falco libs client might require a kernel module name which is not called `falco` / `falco`, depending on the client. Allow these two options to be passed as command line flags via `--moduledrivername` and `--moduledevicename`

Fixes #

```release-note
new(driverkit): add --moduledrivername and --moduledevicename to change the name of the built kernel module.
```

### DIFF
--- a/cmd/root_options.go
+++ b/cmd/root_options.go
@@ -21,8 +21,8 @@ type RootOptions struct {
 	Architecture     string `default:"x86_64" validate:"required,oneof=x86_64" name:"architecture"`
 	DriverVersion    string `default:"dev" validate:"eq=dev|sha1|semver" name:"driver version"`
 	KernelVersion    uint16 `default:"1" validate:"omitempty,number" name:"kernel version"`
-	ModuleDriverName string `default:"falco" name:"kernel module driver name"`
-	ModuleDeviceName string `default:"falco" name:"kernel module device name"`
+	ModuleDriverName string `default:"falco" validate:"max=60" name:"kernel module driver name"`
+	ModuleDeviceName string `default:"falco" validate:"excludes=/,max=255" name:"kernel module device name"`
 	KernelRelease    string `validate:"required,ascii" name:"kernel release"`
 	Target           string `validate:"required,target" name:"target"`
 	KernelConfigData string `validate:"omitempty,base64" name:"kernel config data"` // fixme > tag "name" does not seem to work when used at struct level, but works when used at inner level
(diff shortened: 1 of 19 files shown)


pkg/driverbuilder/builder/builders.go
```diff
@@ -12,16 +12,16 @@ import (
 const DriverDirectory = "/tmp/driver"
 
 // ModuleFileName is the standard file name for the kernel module.
-const ModuleFileName = "falco.ko"
+const ModuleFileName = "module.ko"
```
leodido, line 15: I'm not getting why this change, Luca.

Not that it is a huge problem, the patch looks good!

leodido (commented):
Please update the README accordingly, too

Ty!

pkg/driverbuilder/builder/builders.go
LucaGuerra: It isn't really necessary, but I wanted to catch cases where the `falco` string is hardcoded so when running the script in the container it wouldn't "accidentally" work. Also, it's symmetrical with `probe.o`. No big deal either way though, let me know if you feel like it's better to leave it `falco` and I'll revert this change.

LucaGuerra:
Thanks a lot for the review @leodido ! Good catch on the readme, I added a commit for the readme.

pkg/driverbuilder/builder/builders.go
leodido: Naaa, it makes sense. I understood your reasoning now.

cmd/root_options.go
```diff
@@ -21,6 +21,8 @@ type RootOptions struct {
 	Architecture     string `default:"x86_64" validate:"required,oneof=x86_64" name:"architecture"`
 	DriverVersion    string `default:"dev" validate:"eq=dev|sha1|semver" name:"driver version"`
 	KernelVersion    uint16 `default:"1" validate:"omitempty,number" name:"kernel version"`
+	ModuleDriverName string `default:"falco" name:"kernel module driver name"`
```
leodido, line 24: Worth adding a validator to restrict the alphabet a driver/device name can have?

Not sure any character is possible...

leodido (commented):
Left another comment :)

cmd/root_options.go
LucaGuerra: Sorry about the delay. Anyways I went down that specific rabbit hole and it looks to me that:
* A kernel driver does not seem to have specific name restriction if not a max length of 60 characters ( https://elixir.bootlin.com/linux/latest/source/include/linux/moduleparam.h#L21 ). In fact, I tried editing the elf and changing it to `falco🦅` and it does work, although I wouldn't recommend it.
* The device is just a file name, so max 255 and no slash should be enough

I added a commit with these rules

leodido: Worth checking, innit?

leodido (approved):

PR #122, opened 2022-01-19
new(tests): Add outContains for tests

LucaGuerra:
Signed-off-by: Luca Guerra <luca@guerra.sh>

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area build

> /area cmd

> /area pkg

> /area docs

/area tests

The CLI tests are useful to check for many errors and issues. However, every time we add a new parameter or modify an existing one there are six test fixtures to change, a number that may be growing depending on the tests that we add.
I'd like to propose introducing a way to only check a part of the output so we don't necessarily have to copy paste the usage message several times when we have to, say, simply check an error condition.
Regardless, two tests are still left that exercise the code that generates the usage message.

Fixes #

```release-note

```


leodido (changes requested):
The idea is good.

I would not remove the existing test cases/fixtures, tho. I'm contrary to that because it's useful to test the whole CLI.

Anyways, since I understand where you're coming from and it's reasonable, I suggest you further extend this PR implementing a way to tell the tests to only execute the "contains" matching rather than the full match one.

Then, in the tests, we can use this modality. But power users and maintainers still would have the option (by using the extended match modality) to check everything in the CLI is fully working as designed and intended.

LucaGuerra:
Thanks for the feedback @leodido .

I believe that unit tests are extremely useful and they make sense only if they're executed automatically. This is why I strongly support running every test for every PR. I think I see your point about allowing maintainers and contributors to run the full check for extra verification purposes, but I feel that those will diverge from the actual code unless they're ran automatically at every PR, which is what we have now.

Perhaps what I'm thinking here is that I don't see the added value of checking the entire message for each test vs only one part. Essentially, keeping the tests as-is would check if the usage message is printed incorrectly in some error cases when the error is itself correct and I don't think it's worth to have contributors copy-paste the message several times just to enable this additional check.

I was thinking about another option: we could add instead a field `shouldDisplayUsage` that makes sure that the tests checks that the output contains the usage message verbatim. This would save contributors from locating all the test fixtures that use it but still exercise the case mentioned above. What do you think?

FedeDP:
> Perhaps what I'm thinking here is that I don't see the added value of checking the entire message for each test vs only one part. Essentially, keeping the tests as-is would check if the usage message is printed incorrectly in some error cases when the error is itself correct and I don't think it's worth to have contributors copy-paste the message several times just to enable this additional check.

IMHO, i agree with Luca on this one. I was shocked when i saw that i needed to paste the same lines over and over iterating over N different files.
From the contributor pov it makes no sense, other than losing like 10minutes (oh, and forgetting about it every time :D)

dwindsor:
> I was thinking about another option: we could add instead a field shouldDisplayUsage that makes sure that the tests checks that the output contains the usage message verbatim. This would save contributors from locating all the test fixtures that use it but still exercise the case mentioned above. What do you think?

Rather than adding `shouldDisplayUsage` for just the usage message case, why not implement the suggestion that @leodido made to enable a "contains" mode for tests in general? I assume there are other types of tests that don't need full match capability.

LucaGuerra:
Thanks @dwindsor for the comment! I'd like to see if I understood the suggestions correctly:
* My original patch was meant to add an "outContains" method to tests that would only check if the output of a command contains a specific string. If it does, the test passes. One thing that this can be used to (in my opinion) improve tests would be to remove copy-pasted outputs from the usage message to improve contributor experience when a command line option is changed.
* Maintainers requested changes on that. The suggestion, if I understood it right, is to have the same test check for both contains and full match, but by default run only the contains part. I'm not very happy about that because that would mean that the full output part would be quickly outdated without people realizing (because by default the full output is not checked).

However, by reading your comment, I believe you understood the right suggestion better than I did :sweat_smile: . What would be the right way in your opinion? I agree that having a way of partially checking output makes sense and I'll happily implement the suggestion if I get it right!

FedeDP:
Hi all!
I think https://github.com/falcosecurity/driverkit/pull/193 can be the best solution, leveraging text.template package (that is the same we use for builder's template data) to execute predefined, dynamic templates on cmd package test txt files.
Please have a look :pray:

LucaGuerra:
I agree that using templates solves the underlying issue!

PR #124, opened 2022-02-12
feat(pkg/kernelrelease): introduce json tags for kernelrelease

maxgio92:
This PR introduces JSON tags for the `KernelRelease` struct, with snake case naming convention.

Signed-off-by: maxgio92 <massimiliano.giovagnoli.1992@gmail.com>

**What type of PR is this?**

/kind feature

/area pkg
/area tests

If we'd need to encode `KernelRelease`s the transformed data will follow the snake case naming convention.

None

This PR introduces a test for the struct encoding transformation.

```release-note
feat(pkg/kernelrelease): introduce json tags for kernelrelease
action required: KernelRelease JSON-encoded data fields are renamed following snake case
```

### DIFF
--- a/pkg/kernelrelease/kernelrelease.go
+++ b/pkg/kernelrelease/kernelrelease.go
@@ -11,12 +11,12 @@ var (
 
 // KernelRelease contains all the version parts.
 type KernelRelease struct {
-	Fullversion      string
-	Version          string
-	PatchLevel       string
-	Sublevel         string
-	Extraversion     string
-	FullExtraversion string
+	Fullversion      string `json:"full_version"`
+	Version          string `json:"version"`
+	PatchLevel       string `json:"patch_level"`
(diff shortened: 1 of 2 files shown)


leodido (commented):
Hello Max, thanks for this.

Anyways, I'm struggling to understand the actual need for this.

Can you expand a bit on it?

I also noticed you are using "action requested" in the release notes block. Do you will to communicate a breaking change?

maxgio92:
Hey @leodido, I'd like to leverage this package for scraping of kernel releases. So I'd like to manage these objects as KernelRelease objects, and render the output in JSON.
I'd like that the fields follow snake case convention.
That's it :)

What do you think?

Edit: Maybe the test is not needed as it's actually testing JSON Marshal().

maxgio92:
About the action required I imaged that if consumers already marshalled this type of data, would need to take care of this fields renaming.

PR #125, opened 2022-02-16
fix(build): support PPM_API_CURRENT_VERSION macros

LucaGuerra:
Signed-off-by: Luca Guerra <luca@guerra.sh>

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/area build

> /area cmd

> /area pkg

> /area docs

> /area tests

This PR adds support for scap/falco driver versioning, introduced in https://github.com/falcosecurity/libs/pull/39 . Currently, driverkit is broken on any version equal or newer than the above.
Since Driverkit does not use CMake to generate the driver header file, we need to do it at runtime by following the same process that the cmake generator does.

Fixes #

```release-note
fix(build): add support for Falco/Scap driver API versioning macros.
```

### DIFF
--- a/pkg/driverbuilder/templates.go
+++ b/pkg/driverbuilder/templates.go
@@ -74,6 +74,9 @@ type driverConfigData struct {
 
 // XXX both PROBE and DRIVER variables are kept for now so that Driverkit is compatible with older versions.
 // they can be removed when versions from early 2022/late 2021 will not be supported anymore.
+
+// Note that in the future DRIVER_COMMIT will be different from DRIVER_VERSION. Currently, it is the same as the commit
+// and no decision has been made yet about the distinction in falcosecurity/libs. Will need to be updated.
 const fillDriverConfigTemplate = `
 set -euxo pipefail
 
--- a/pkg/driverbuilder/templates.go
+++ b/pkg/driverbuilder/templates.go
@@ -72,25 +72,56 @@ type driverConfigData struct {
 	DeviceName    string
 }
 
-// TODO(fntlnz): these variables are still called probes because of how the driver code
-// references to them. We must change the driver code and then change here to s/PROBE_*/MODULE_*/g
-const driverConfigTemplate = `
+const fillDriverConfigTemplate = `
+set -euxo pipefail
+
+DRIVER_BUILD_DIR=$1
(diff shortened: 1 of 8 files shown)


pkg/driverbuilder/templates.go
```diff
@@ -72,25 +72,60 @@ type driverConfigData struct {
...
+DRIVER_BUILD_DIR=$1
+DRIVER_CONFIG_FILE="$DRIVER_BUILD_DIR/driver_config.h"
+
+cat << EOF > $DRIVER_CONFIG_FILE
 #pragma once
 
-#define PROBE_VERSION "{{ .DriverVersion }}"
+#define DRIVER_VERSION "{{ .DriverVersion }}"
+
+#define DRIVER_COMMIT "{{ .DriverVersion }}"
```
leodido, line 88: Why this? Where it is used?

leodido (commented):
Left a question

pkg/driverbuilder/templates.go
FedeDP: https://github.com/falcosecurity/libs/blob/cc42b767e854256355a25f5a3779114b63bd056c/driver/driver_config.h.in

It entered libs with https://github.com/falcosecurity/libs/pull/39!

leogr: Hey @leodido 

note that `DRIVER_COMMIT` is intended only for debugging purposes. 

@LucaGuerra please note that in the future `DRIVER_COMMIT` will be different from `DRIVER_VERSION`.
For example `DRIVER_COMMIT` will continue to be equal to the commit id, but  `DRIVER_VERSION` will be equal to the driver version (es 1.0.0).

LucaGuerra: Thanks! Yes, I really didn't know what to put in `DRIVER_VERSION`. I expect that could change in the future but we can't predict it right now I think.

leodido: Thank you, now it makes sense.

It's probably worth leaving a comment or opening an issue to remember the distinction...

LucaGuerra: ... and the comment was added!

leodido (approved):

PR #128, opened 2022-03-09
fix(driverkit): Bump binutils version, install zstd in base container

### LINKED ISSUE #127: Driverkit fails to build Ubuntu releases because it's missing zstd
When trying to build a driver for newer Ubuntu kernels, `driverkit` fails to extract the downloaded kernel headers package because `zstd` is missing:

```
DEBU + curl --silent -o kernel.deb -SL https://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux/linux-headers-5.13.0-30_5.13.0-30.33_all.deb
DEBU + ar x kernel.deb
DEBU + tar -xvf data.tar.zst
DEBU mtar (child): zstd: Cannot exec: No such file or directory
DEBU tar (child): Error is not recoverable: exiting now
DEBU Htar: Child returned status 2
DEBU tar: Error is not recoverable: exiting now
```

Adding `zstd` to the `driverkit` base container image isn't possible because it requires a version of `binutils` that's too recent:

```
root@c660c81620df:/# apt install zstd
Reading package lists... Done
Building dependency tree
Reading state information... Done
You might want to run 'apt --fix-broken install' to correct these.
The following packages have unmet dependencies:
 gcc-8 : Depends: binutils (>= 2.31.1) but 2.30-22 is to be installed
 ```
 
 To fix this, we need bump the version of `binutils` currently hosted on `download.falco.org/dependencies` to be the required version (>= `2.31.1`) and then add `zstd` to `build/builder.Dockerfile`.

### PR DESCRIPTION
**Pull request by dwindsor**
Signed-off-by: David Windsor <dwindsor@secureworks.com>
Signed-off-by: Logan Bond <lbond@secureworks.com>

**What type of PR is this?**

/kind bug

/area build

When trying to build a driver for newer Ubuntu kernels, `driverkit` fails to extract the downloaded kernel headers package because `zstd` is missing. But, `zstd` requires a version of `binutils` that's more recent than the one on [https://download.falco.org/dependencies](https://download.falco.org/dependencies/), so we need to both:

1) Bump the version of `binutils` hosted on [https://download.falco.org/dependencies](https://download.falco.org/dependencies/) to be >= `2.31.1`
2) Add `zstd` to `build/builder.Dockerfile`

Fixes #127 

**NOTE:** We're unable to upload files to download.falco.org, so the URLs to the updated binutils packages will need to be updated by a maintainer after the exact filename has been determined.

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ IMAGE_NAME_DRIVERKIT_REF := $(IMAGE_NAME_DRIVERKIT):$(GIT_REF)
 IMAGE_NAME_DRIVERKIT_COMMIT := $(IMAGE_NAME_DRIVERKIT):$(GIT_COMMIT)
 IMAGE_NAME_DRIVERKIT_LATEST := $(IMAGE_NAME_DRIVERKIT):latest
 
-LDFLAGS := -X github.com/falcosecurity/driverkit/pkg/version.buildTime=$(shell date +%s) -X github.com/falcosecurity/driverkit/pkg/version.gitCommit=${GIT_COMMIT} -X github.com/falcosecurity/driverkit/pkg/version.gitTag=$(if ${GIT_TAG},${GIT_TAG},v0.0.0) -X github.com/falcosecurity/driverkit/pkg/version.commitsFromGitTag=${COMMITS_FROM_GIT_TAG} -X github.com/falcosecurity/driverkit/pkg/driverbuilder.builderBaseImage=${IMAGE_NAME_BUILDER_COMMIT}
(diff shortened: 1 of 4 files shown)


leodido (commented):
Thanks for this @dwindsor!

At first sight (haven't tested it out) it looks good to me (yes, the binutils URL is to be changed).

I'm just wondering whether it impacts or it has side effects on other targets (different from Ubuntu).

I haven't tested it out, but maybe you have. Does it? :)

dwindsor:
> Thanks for this @dwindsor!
> 
> At first sight (haven't tested it out) it looks good to me (yes, the binutils URL is to be changed).
> 
> I'm just wondering whether it impacts or it has side effects on other targets (different from Ubuntu).
> 
> I haven't tested it out, but maybe you have. Does it? :)

Sure! We did do some testing, but admittedly just a bit. We were able to build a new CentOS 7 driver after making this change.

PR #130, opened 2022-03-24
Add driverbuilder for Rocky Linux

dwindsor:
Signed-off-by: David Windsor <dwindsor@secureworks.com>

```bash

./_output/bin/driverkit -l debug docker --output-module \ /tmp/falco_rocky.ko --kernelversion=1
--kernelrelease=4.18.0-348.el8.0.2.x86_64 --driverversion=b7eb0dd65226a8dc254d228c8d950d07bf3521d2
--moduledrivername="falco_rocky" --moduledevicename="falco-rocky" --target=rocky

DEBU Ifilename:       /tmp/driver/module.ko 
DEBU schema_version: 1.0.0                        
DEBU api_version:    1.0.0                        
DEBU build_commit:   b7eb0dd65226a8dc254d228c8d950d07bf3521d2 
DEBU version:        b7eb0dd65226a8dc254d228c8d950d07bf3521d2 
DEBU author:         the Falco authors            
DEBU license:        GPL                          
DEBU rhelversion:    8.5                          
DEBU srcversion:     FB3CE3C01854B85C4CCF472      
DEBU depends:                                     
DEBU name:           falco_rocky                  
DEBU vermagic:       4.18.0-348.el8.0.2.x86_64 SMP mod_unload modversions  
DEBU parm:           max_consumers:Maximum number of consumers that can simultaneously open the devices (uint) 
DEBU parm:           verbose:Enable verbose logging (bool) 

```

**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test
/kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:
/area build

> /area cmd

> /area pkg

> /area docs

> /area tests

**What this PR does / why we need it**:

CentOS was EOL'ed in December, 2021. Rocky Linux is the successor to CentOS. Add support to driverkit for building Rocky Linux kernel drivers.

**Which issue(s) this PR fixes**:

Fixes #129 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
new: support for Rocky Linux
```

PR #131, opened 2022-04-01
Add builder for Flatcar

jepio:
**Example output**
```
# Requires the builderimage to be updated, so replicating this might need also a --builderimage argument
./_output/bin/driverkit docker -l debug -t flatcar --kernelrelease 3255.0.0  --output-module=/tmp/falco.ko --driverversion master

DEBU running without a configuration file         
DEBU running with options                          arch=amd64 driverversion=master kernelrelease=3255.0.0 kernelversion=1 output-module=/tmp/falco.ko target=flatcar
DEBU doing a new docker build                     
DEBU kernel header url found                       url="https://alpha.release.flatcar-linux.net/amd64-usr/3255.0.0/flatcar_production_image_packages.txt"
DEBU kernel header url found                       url="https://alpha.release.flatcar-linux.net/amd64-usr/3255.0.0/flatcar_production_image_kernel_config.txt"
DEBU kernel header url found                       url="https://cdn.kernel.org/pub/linux/kernel/v5.x/linux-5.15.43.tar.xz"
...
DEBU make[1]: Leaving directory '/tmp/kernel' 
DEBU + mv falco.ko /tmp/driver/module.ko  
DEBU + strip -g /tmp/driver/module.ko     
DEBU + modinfo /tmp/driver/module.ko      
DEBU filename:       /tmp/driver/module.ko 
DEBU schema_version: 1.0.0                        
DEBU api_version:    1.0.0                        
DEBU build_commit:   master                       
DEBU version:        master                       
DEBU author:         the Falco authors            
DEBU license:        GPL                          
DEBU srcversion:     F2DE132E5F90A47892185E9      
DEBU depends:                                     
DEBU retpoline:      Y                            
DEBU name:           falco                        
DEBU vermagic:       5.15.43-flatcar SMP mod_unload  
DEBU parm:           max_consumers:Maximum number of consumers that can simultaneously open the devices (uint) 
DEBU parm:           verbose:Enable verbose logging (bool) 
INFO kernel module available                       path=/tmp/falco.ko
DEBU context canceled
```

**What type of PR is this?**
/kind feature

/area build

This PR introduces a builder for Flatcar. There have been several reports of vanilla builder not working well for Flatcar (https://github.com/falcosecurity/driverkit/issues?q=flatcar), we also have reports on our own issue tracker of the stock Falco build process not working since we shipped a newer GLIBC (https://github.com/flatcar-linux/Flatcar/issues/534). In slack there are occasional questions in slack regarding the kernel module build for Flatcar. So we know that there are plenty of users that would like this to work.
 
Example output:

 
Fixes #

I tried to not introduce an extra "osrelease" or "osversion" field but that's actually what we would need. I snuck the "osrelease" into the "kernelrelease" field instead, hope that's ok. Also: any thoughts on getting a GCC newer than 8 into the builder container?

```release-note
Add support for Flatcar Container Linux
```

### DIFF
--- a/pkg/driverbuilder/builder/flatcar.go
+++ b/pkg/driverbuilder/builder/flatcar.go
@@ -32,7 +32,7 @@ func (c flatcar) Script(cfg Config) (string, error) {
 	}
 
 	kr := kernelrelease.FromString(cfg.Build.KernelRelease)
-	if kr.Extraversion != "flatcar" {
+	if kr.Extraversion != "" {
 		return "", fmt.Errorf("unexpected extraversion: %s", kr.Extraversion)
 	}
 
--- a/pkg/driverbuilder/builder/flatcar.go
+++ b/pkg/driverbuilder/builder/flatcar.go
@@ -31,7 +31,7 @@ func (c flatcar) Script(cfg Config) (string, error) {
 		return "", err
 	}
 
-	kr := kernelrelease.FromString(cfg.Build.KernelRelease)
+	kr :=  kernelReleaseFromBuildConfig(cfg.Build)
 	if kr.Extraversion != "" {
 		return "", fmt.Errorf("unexpected extraversion: %s", kr.Extraversion)
 	}
@@ -45,12 +45,12 @@ func (c flatcar) Script(cfg Config) (string, error) {
 		return "", fmt.Errorf("not a valid flatcar release version: %s", kr.Version)
 	}
 	flatcarVersion := kr.Fullversion
-	flatcarInfo, err := fetchFlatcarMetadata(flatcarVersion)
+	flatcarInfo, err := fetchFlatcarMetadata(kr)
 	if err != nil {
 		return "", err
 	}
 
-	kconfUrls, err := getResolvingURLs(fetchFlatcarKernelConfigURL(flatcarInfo.Channel, kr.Fullversion))
+	kconfUrls, err := getResolvingURLs(fetchFlatcarKernelConfigURL(kr.Architecture, flatcarInfo.Channel, kr.Fullversion))
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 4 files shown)


pkg/driverbuilder/builder/flatcar.go
```diff
@@ -0,0 +1,232 @@
...
+mkdir -p /tmp/kernel
+mv /tmp/kernel-download/*/* /tmp/kernel
+
+# Change current gcc
+ln -sf /usr/bin/gcc-{{ .GCCVersion }} /usr/bin/gcc
+
+curl --silent -o /tmp/kernel.config -SL {{ .KernelConfigURL }}
+
+cd /tmp/kernel
+make KCONFIG_CONFIG=/tmp/kernel.config oldconfig
```
dwindsor, line 201: I wonder if these calls to make `oldconfig`, `modules_prepare` are necessary? I don't actually see them in any of the other builders, but this may be required for flatcar?

jepio: I copied this part from the vanilla builder. Our own "kernel-devel" package equivalent is compiled against newer glibc, such that it doesn't work with the old debian based builder image. Here we're pulling generic kernel sources (we don't patch our sources) and need to "prepare" those.

jepio:
@leodido any chance you could have a look at this?

dwindsor:
Hi! This looks fine to me. kernel-crawler already has [support](https://github.com/falcosecurity/kernel-crawler/blob/main/kernel_crawler/flatcar.py) for Flatcar, so it makes sense to add this.

dwindsor:
> Thanks for the link to kernel-crawler, I checked what output it produces, and decided to drop the -flatcar from the kernelrelease value to be consistent with it. None of the package naming/formats/locations have changed, but I did catch 2 minor issues that I fixed: I was missing setting EXTRAVERSION correctly and we need dwarves/pahole in the builder container.

Nice! Thanks for the updates. Have you had a chance to try to load the built drivers/probes on Flatcar yet?

jepio:
Yes, that's how I came across the issues 2 issues I found. The module now loads correctly on flatcar.

dwindsor:
Could you update the description with an example command-line invocation building a flatcar driver with successful results as in [PR130](https://github.com/falcosecurity/driverkit/pull/130):

Screen Shot 2022-06-03 at 1 40 20 PM

PR #161, opened 2022-06-16
More Ubuntu Support

### LINKED ISSUE #144: Properly support more ubuntu flavours
**Motivation**

[Kernel crawler](https://github.com/falcosecurity/kernel-crawler) discovered various ubuntu flavours that are unsupported at the moment:
* ubuntu-oracle
* ubuntu-kvm
* ubuntu-oem
* ubuntu-gcp
* ubuntu-azure
* ubuntu-hwe

Moreover, alongside those, there are some more that have the kernel-version suffix, like:
* ubuntu-azure-4.15
* ubuntu-aws-5.0
* ubuntu-aws-5.4

And many more.

**Feature**

We should support all of them.

**Alternatives**

We can just support a generic `ubuntu` target, and force `kernelurls` option (introduced by #143) to be passed.

Given how complex would it be to support any ubuntu flavour, and given that searching for the proper urls takes a human like 2minutes, i think it would not be too bad to enforce the usage of `kernelurls` for it.  

Automatically generated config files from kernel-crawler output will contain the new option in any case, when driverkit will support it.

I think that enforcing kernelurls param to be passed is not as bad as it sounds; again, it would greatly enlarge our supported matrix with like a 2mins effort by the contributors.

**Additional context**

EXONER4TED:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/area build

> /area cmd

> /area pkg

> /area docs

> /area tests

Trying to simplify and expand Driverkit's Ubuntu support:
- Rather than having `ubuntu-generic` and `ubuntu-aws`, simplify to just `ubuntu`
- Expand support to more than just `generic` and `aws`, as described in: #144 

Fixes #144 

This work is also built in part on top of the work @amalone-scwx did in some internal testing - wanted to make sure I give credit where credit is due :)

Yes, simplifying Ubuntu target types to just `--target ubuntu`, though backwards compatibility with `ubuntu-aws` and `ubuntu-generic` should still work (but I would argue need to be deprecated in the future). 

The one "gotcha" to look out for with this change is that now the Ubuntu "Flavor" (ex: aws, generic, kvm, oracle, etc) is now dynamically parsed from the `kernelrelease` field. If the `kernelrelease` passed in is formatted incorrectly or omits that flavor, then `generic` is the assumed flavor.

For example, if a config looks like:
```
kernelversion: 199
kernelrelease: 4.15.0-188
target: ubuntu
```
The updated code will assume that `generic` is the "flavor" here since the `target` is no longer a reliable way to get that information.

```release-note
Simplifying and expanding Ubuntu support into just `--target ubuntu` and adding support for all flavors! `ubuntu-generic` and `ubuntu-aws` target types are still supported for backwards compatibility, but should be deprecated.
Massive overhaul to the go-side of the Ubuntu builder which now parses the Ubuntu "Flavor" from the kernelrelease and uses it to build URL paths to search for kernel headers, meaning all Ubuntu flavors should now be supported rather than just AWS and generic.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ IMAGE_NAME_DRIVERKIT_REF := $(IMAGE_NAME_DRIVERKIT):$(GIT_REF)
 IMAGE_NAME_DRIVERKIT_COMMIT := $(IMAGE_NAME_DRIVERKIT):$(GIT_COMMIT)
 IMAGE_NAME_DRIVERKIT_LATEST := $(IMAGE_NAME_DRIVERKIT):latest
 
-LDFLAGS := -X github.com/falcosecurity/driverkit/pkg/version.buildTime=$(shell date +%s) -X github.com/falcosecurity/driverkit/pkg/version.gitCommit=${GIT_COMMIT} -X github.com/falcosecurity/driverkit/pkg/version.gitTag=$(if ${GIT_TAG},${GIT_TAG},v0.0.0) -X github.com/falcosecurity/driverkit/pkg/version.commitsFromGitTag=${COMMITS_FROM_GIT_TAG} -X github.com/falcosecurity/driverkit/pkg/driverbuilder.BuilderBaseImage=${IMAGE_NAME_BUILDER_COMMIT}
+LDFLAGS := -X github.com/falcosecurity/driverkit/pkg/version.buildTime=$(shell date +%s) -X github.com/falcosecurity/driverkit/pkg/version.gitCommit=${GIT_COMMIT} -X github.com/falcosecurity/driverkit/pkg/version.gitTag=$(if ${GIT_TAG},${GIT_TAG},v0.0.0) -X github.com/falcosecurity/driverkit/pkg/version.commitsFromGitTag=${COMMITS_FROM_GIT_TAG} -X github.com/falcosecurity/driverkit/pkg/driverbuilder.BuilderBaseImage=${IMAGE_NAME_BUILDER_LATEST}
 
 ARCHS := "arm64,amd64"
 
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ IMAGE_NAME_DRIVERKIT_REF := $(IMAGE_NAME_DRIVERKIT):$(GIT_REF)
 IMAGE_NAME_DRIVERKIT_COMMIT := $(IMAGE_NAME_DRIVERKIT):$(GIT_COMMIT)
 IMAGE_NAME_DRIVERKIT_LATEST := $(IMAGE_NAME_DRIVERKIT):latest
 
-LDFLAGS := -X github.com/falcosecurity/driverkit/pkg/version.buildTime=$(shell date +%s) -X github.com/falcosecurity/driverkit/pkg/version.gitCommit=${GIT_COMMIT} -X github.com/falcosecurity/driverkit/pkg/version.gitTag=$(if ${GIT_TAG},${GIT_TAG},v0.0.0) -X github.com/falcosecurity/driverkit/pkg/version.commitsFromGitTag=${COMMITS_FROM_GIT_TAG} -X github.com/falcosecurity/driverkit/pkg/driverbuilder.BuilderBaseImage=${IMAGE_NAME_BUILDER_LATEST}
+LDFLAGS := -X github.com/falcosecurity/driverkit/pkg/version.buildTime=$(shell date +%s) -X github.com/falcosecurity/driverkit/pkg/version.gitCommit=${GIT_COMMIT} -X github.com/falcosecurity/driverkit/pkg/version.gitTag=$(if ${GIT_TAG},${GIT_TAG},v0.0.0) -X github.com/falcosecurity/driverkit/pkg/version.commitsFromGitTag=${COMMITS_FROM_GIT_TAG} -X github.com/falcosecurity/driverkit/pkg/driverbuilder.BuilderBaseImage=${IMAGE_NAME_BUILDER_COMMIT}
 
 ARCHS := "arm64,amd64"
 
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ integration_test: $(test_configs)
 
 .PHONY: $(test_configs)
 $(test_configs): ${driverkit}
-	${driverkit} docker -c $@ --builderimage falcosecurity/driverkit-builder:latest -l debug --timeout 600
+	${driverkit} docker -c $@ --builderimage $(IMAGE_NAME_BUILDER_COMMIT) -l debug --timeout 600
 
 .PHONY: ${driverkit_docgen}
 ${driverkit_docgen}: ${PWD}/docgen
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ push/latest:
 .PHONY: test
 test:
 	go test -v -cover -race ./...
-	go test -v -buildmode=pie ./cmd
+	go test -v -cover -buildmode=pie ./cmd
 
 .PHONY: integration_test
 integration_test: $(test_configs)
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ integration_test: $(test_configs)
 
 .PHONY: $(test_configs)
 $(test_configs): ${driverkit}
-	${driverkit} docker -c $@ --builderimage $(IMAGE_NAME_BUILDER_COMMIT) -l debug --timeout 600
+	${driverkit} docker -c $@ --builderimage falcosecurity/driverkit-builder:latest -l debug --timeout 600
 
 .PHONY: ${driverkit_docgen}
 ${driverkit_docgen}: ${PWD}/docgen
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,7 @@ push/latest:
 
 .PHONY: test
 test:
-	go test -v -race ./...
+	go test -v -cover -race ./...
 	go test -v -buildmode=pie ./cmd
 
 .PHONY: integration_test
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -37,34 +37,34 @@ func persistentValidateFunc(rootCommand *RootCmd, rootOpts *RootOptions) func(c 
 			"output-probe":  "output.probe",
 		}
 		rootCommand.c.Flags().VisitAll(func(f *pflag.Flag) {
-			if name := f.Name; !skip[name] {
-				if name == "kernelurls" {
-					// Slice types need special treatment when used as flags. If we call 'Set(name, value)',
-					// rather than replace, it appends. Since viper will already have the cli options set
-					// if supplied, we only need this step if rootCommand doesn't already have them e.g.
-					// not set on CLI so read from config.
-					if cli_urls, err := rootCommand.c.Flags().GetStringSlice(name); err == nil && len(cli_urls) != 0 {
-						return
-					}
-					value := viper.GetStringSlice(name)
-					if len(value) != 0 {
-						strValue := strings.Join(value, ",")
-						rootCommand.c.Flags().Set(name, strValue)
-					}
-				} else {
-					value := viper.GetString(name)
-					if value == "" {
-						// fallback to nested options in config file, if any
-						if nestedName, ok := nested[name]; ok {
-							value = viper.GetString(nestedName)
-						}
-					}
-					// set the value, if any, otherwise let the default
-					if value != "" {
-						rootCommand.c.Flags().Set(name, value)
-					}
-				}
-			}
+		    if name := f.Name; !skip[name] {
+                if name == "kernelurls" {
+                    // Slice types need special treatment when used as flags. If we call 'Set(name, value)',
+                    // rather than replace, it appends. Since viper will already have the cli options set
+                    // if supplied, we only need this step if rootCommand doesn't already have them e.g.
+                    // not set on CLI so read from config.
+                    if cli_urls, err := rootCommand.c.Flags().GetStringSlice(name); err == nil && len(cli_urls) != 0 {
+                       return
+                    }
+                    value := viper.GetStringSlice(name)
+                    if len(value) != 0 {
+                        strValue := strings.Join(value, ",")
+                        rootCommand.c.Flags().Set(name, strValue)
+                    }
+                } else {
+                    value := viper.GetString(name)
+                    if value == "" {
+                        // fallback to nested options in config file, if any
+                        if nestedName, ok := nested[name]; ok {
+                            value = viper.GetString(nestedName)
+                        }
+                    }
+                    // set the value, if any, otherwise let the default
+                    if value != "" {
+                        rootCommand.c.Flags().Set(name, value)
+                    }
+                }
+            }
 		})
 
 		// Avoid sensitive info into default values help line
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -37,34 +37,34 @@ func persistentValidateFunc(rootCommand *RootCmd, rootOpts *RootOptions) func(c 
 			"output-probe":  "output.probe",
 		}
 		rootCommand.c.Flags().VisitAll(func(f *pflag.Flag) {
-		    if name := f.Name; !skip[name] {
-                if name == "kernelurls" {
-                    // Slice types need special treatment when used as flags. If we call 'Set(name, value)',
-                    // rather than replace, it appends. Since viper will already have the cli options set
-                    // if supplied, we only need this step if rootCommand doesn't already have them e.g.
-                    // not set on CLI so read from config.
-                    if cli_urls, err := rootCommand.c.Flags().GetStringSlice(name); err == nil && len(cli_urls) != 0 {
-                       return
-                    }
-                    value := viper.GetStringSlice(name)
-                    if len(value) != 0 {
-                        strValue := strings.Join(value, ",")
-                        rootCommand.c.Flags().Set(name, strValue)
-                    }
-                } else {
-                    value := viper.GetString(name)
-                    if value == "" {
-                        // fallback to nested options in config file, if any
-                        if nestedName, ok := nested[name]; ok {
-                            value = viper.GetString(nestedName)
-                        }
-                    }
-                    // set the value, if any, otherwise let the default
-                    if value != "" {
-                        rootCommand.c.Flags().Set(name, value)
-                    }
-                }
-            }
+			if name := f.Name; !skip[name] {
+				if name == "kernelurls" {
+					// Slice types need special treatment when used as flags. If we call 'Set(name, value)',
+					// rather than replace, it appends. Since viper will already have the cli options set
+					// if supplied, we only need this step if rootCommand doesn't already have them e.g.
+					// not set on CLI so read from config.
+					if cli_urls, err := rootCommand.c.Flags().GetStringSlice(name); err == nil && len(cli_urls) != 0 {
+						return
+					}
+					value := viper.GetStringSlice(name)
+					if len(value) != 0 {
+						strValue := strings.Join(value, ",")
+						rootCommand.c.Flags().Set(name, strValue)
+					}
+				} else {
+					value := viper.GetString(name)
+					if value == "" {
+						// fallback to nested options in config file, if any
+						if nestedName, ok := nested[name]; ok {
+							value = viper.GetString(nestedName)
+						}
+					}
+					// set the value, if any, otherwise let the default
+					if value != "" {
+						rootCommand.c.Flags().Set(name, value)
+					}
+				}
+			}
 		})
 
 		// Avoid sensitive info into default values help line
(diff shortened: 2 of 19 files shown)


FedeDP:
Hi! This is great!
My only concern is dropping backward compatibility for existing configs with `ubuntu-aws`.
I think we should actually support `ubuntu-aws` from cmdline/config and internally store it as 'ubuntu`.
Aside from that, i think this needs a lot of testing before being merged :) 
But it's a great step forward IMHO; thank you very much!

EXONER4TED:
That makes sense to me! I can make sure to add in that backwards compatibility. I'm unable to continue work on this PR for a few days, but I'll get that going as soon as I get back to it.

Makefile
```diff
@@ -28,7 +28,7 @@ IMAGE_NAME_DRIVERKIT_REF := $(IMAGE_NAME_DRIVERKIT):$(GIT_REF)
 IMAGE_NAME_DRIVERKIT_COMMIT := $(IMAGE_NAME_DRIVERKIT):$(GIT_COMMIT)
 IMAGE_NAME_DRIVERKIT_LATEST := $(IMAGE_NAME_DRIVERKIT):latest
 
-LDFLAGS := -X github.com/falcosecurity/driverkit/pkg/version.buildTime=$(shell date +%s) -X github.com/falcosecurity/driverkit/pkg/version.gitCommit=${GIT_COMMIT} -X github.com/falcosecurity/driverkit/pkg/version.gitTag=$(if ${GIT_TAG},${GIT_TAG},v0.0.0) -X github.com/falcosecurity/driverkit/pkg/version.commitsFromGitTag=${COMMITS_FROM_GIT_TAG} -X github.com/falcosecurity/driverkit/pkg/driverbuilder.BuilderBaseImage=${IMAGE_NAME_BUILDER_COMMIT}
+LDFLAGS := -X github.com/falcosecurity/driverkit/pkg/version.buildTime=$(shell date +%s) -X github.com/falcosecurity/driverkit/pkg/version.gitCommit=${GIT_COMMIT} -X github.com/falcosecurity/driverkit/pkg/version.gitTag=$(if ${GIT_TAG},${GIT_TAG},v0.0.0) -X github.com/falcosecurity/driverkit/pkg/version.commitsFromGitTag=${COMMITS_FROM_GIT_TAG} -X github.com/falcosecurity/driverkit/pkg/driverbuilder.BuilderBaseImage=${IMAGE_NAME_BUILDER_LATEST}
```
FedeDP, line 31: I think this was not a desired committed change :)

cmd/root.go
```diff
@@ -37,34 +37,34 @@ func persistentValidateFunc(rootCommand *RootCmd, rootOpts *RootOptions) func(c
...
-                            value = viper.GetString(nestedName)
-                        }
-                    }
-                    // set the value, if any, otherwise let the default
-                    if value != "" {
-                        rootCommand.c.Flags().Set(name, value)
-                    }
-                }
-            }
+			if name := f.Name; !skip[name] {
```
FedeDP, line 40: Can we checkout master version of this to avoid adding noise to the PR?

cmd/root.go
```diff
@@ -128,7 +128,7 @@ func NewRootCmd() *RootCmd {
 	flags.StringVar(&rootOpts.Output.Probe, "output-probe", rootOpts.Output.Probe, "filepath where to save the resulting eBPF probe")
 	flags.StringVar(&rootOpts.Architecture, "architecture", runtime.GOARCH, "target architecture for the built driver")
 	flags.StringVar(&rootOpts.DriverVersion, "driverversion", rootOpts.DriverVersion, "driver version as a git commit hash or as a git tag")
-	flags.Uint16Var(&rootOpts.KernelVersion, "kernelversion", rootOpts.KernelVersion, "kernel version to build the module for, it's the numeric value after the hash when you execute 'uname -v'")
+	flags.StringVar(&rootOpts.KernelVersion, "kernelversion", rootOpts.KernelVersion, "kernel version to build the module for, it's the numeric value after the hash when you execute 'uname -v'")
```
FedeDP, line 131: While this is the desired solution, i am not sure it won't break backward compatibility. 
We should double check (CI integration tests will do that for us as soon as you rebase though!)

pkg/driverbuilder/builder/ubuntu.go
```diff
@@ -3,111 +3,71 @@ package builder
...
-        return "", err
-    }
-    if len(urls) < 2 {
-        return "", fmt.Errorf("specific kernel headers not found")
-    }
-
 	td := ubuntuTemplateData{
 		DriverBuildDir:       DriverDirectory,
-		ModuleDownloadURL:    moduleDownloadURL(c),
+		ModuleDownloadURL:    fmt.Sprintf("%s/%s.tar.gz", c.DownloadBaseURL, c.Build.DriverVersion),
```
FedeDP, line 67: Isn't this the exact definition of `moduleDownloadURL`?
https://github.com/falcosecurity/driverkit/blob/76daf8787f96e91092cf8b16ba07c12d39173677/pkg/driverbuilder/builder/builders.go#L51
I'd leave old line here.

Makefile
EXONER4TED: This PR is still in draft - I changed this line to use the latest container publicly available to do my dev as I'm not trying to rewrite the container for now. It'll be changed back when I'm not in draft anymore.

cmd/root.go
EXONER4TED: Yep, again, still in draft mode. I haven't looked at the testing yet. I'm not done with making ubuntu work yet.

pkg/driverbuilder/builder/ubuntu.go
EXONER4TED: Yea, I had actually not made this change. It was pulled in from somewhere else. I do need to flip it back.

cmd/root.go
EXONER4TED: I can revert the linting, sure. Will clean it up when I'm not in draft mode.

FedeDP:
Ah i see, sorry we use the "wip" in the PR title to notify that the PR is a wip :) 
So, forgive me if i didn't notice the draft ahah

EXONER4TED:
All good :) I was using the built-in draft mode of Github, which prevents the PR from being merged:
image

But let me edit the title to make it more clear!

cmd/root.go
EXONER4TED: Done and done

EXONER4TED:
I actually think I'm very nearly done - I was able to build ubuntu-oracle, ubuntu-aws, and ubuntu-kvm, for example, on my machine just fine. However, I can't for the life in me figure out why it won't pass the integration tests haha...

On my M1 Mac, I'm able to compile the ubuntu-aws ARM test just fine:
```
╰─❯ _output/bin/driverkit docker --output-module _output/test.ko --target ubuntu --kernelrelease 4.15.0-1129-aws --kernelversion 138 --kernelurls "https://download.falco.org/fixtures/driverkit/aarch64_devel/linux-headers-4.15.0-1129-aws_4.15.0-1129.138_arm64.deb" --kernelurls "https://download.falco.org/fixtures/driverkit/aarch64_devel/linux-aws-headers-4.15.0-1129_4.15.0-1129.138_all.deb" --architecture arm64         
INFO kernel module available                       path=_output/test.ko
```

I was hoping to have this wrapped up today, but until I can figure out the CI errors... might take a few more days.

Current CI errors:
```
DEBU   CC [M]  /tmp/driver/main.o         
DEBU �In file included from <command-line>: 
DEBU ././include/linux/kconfig.h:5:10: fatal error: generated/autoconf.h: No such file or directory 
DEBU  #include <generated/autoconf.h>             
DEBU           ^~~~~~~~~~~~~~~~~~~~~~             
DEBU compilation terminated.              
DEBU Fmake[2]: *** [scripts/Makefile.build:335: /tmp/driver/main.o] Error 1 
DEBU :make[1]: *** [Makefile:1590: _module_/tmp/driver] Error 2 
DEBU Xmake[1]: Leaving directory '/tmp/kernel-download/usr/src/linux-aws-headers-4.15.0-1129' 
DEBU $make: *** [Makefile:7: all] Error 2  
DEBU context canceled                             
FATA exiting                                       error="Error: No such container:path: 4b5da0bf0adc7dd422a48425a407cf6e5ba0aec8193d30a00cf785c9193daa17:/tmp/driver/module.ko"
make: *** [Makefile:91: test/aarch64/configs/ubuntu_aws.yaml] Error 1
```

EXONER4TED:
Here's initial output from my branch building on local. I used kernel-crawler as my input and looped over anything under `Ubuntu`. It has several new flavors!

However, in the Driverkit output, there were 2270 ubuntu targets to build, and I was only able to build ~1100 of them. Most of them seemed to fail for the same reason the CI fails. So I am still investigating that issue, but I think it's the only one left...

[builds.txt](https://github.com/falcosecurity/driverkit/files/9132514/builds.txt)

EXONER4TED:
I think I figured it out - I had changed:
```
		KernelHeadersPattern: "linux-headers*",
```
To be `linux*headers*` at an earlier point in my debugging when I was troubleshooting something. It seems that is now shooting me in the foot because the template script is grabbing the wrong directory for the `include/generated/autoconf.h` with the wildcard.

I put it back, seeing if that fixes my issue.

pkg/driverbuilder/builder/ubuntu.go
```diff
@@ -251,43 +167,26 @@ func fetchUbuntuGenericKernelURL(baseURL string, kr kernelrelease.KernelRelease,
 			kr.Architecture.String(),
 		),
 		fmt.Sprintf(
-			"%s/linux-headers-%s%s-generic_%s-%s.%d_%s.deb",
-			baseURL,
+			"linux-headers-%s-%s-%s_%s-%s.%s_%s.deb",
```
Lowaiz, line 170: During a Ubuntu Azure test, the second and the last pattern that you check were identical, and due to this, the function `getResolvingURLs` return a array with 3 validated URLs, eventually the function `ubuntuHeadersURLFromRelease` return an error because it was expecting exactly 2 values.
Differents ways to fix it:
- Change the `len(urls) == 2` to `len(urls) >= 2` in `ubuntuHeadersURLFromRelease` function
- Checking for duplicated before returning the patterns in `fetchUbuntuKernelURL` function

Config:
```
kernelversion: 46
kernelrelease: 5.4.0-1044-azure
target: ubuntu
architecture: amd64
output:
  module: output/39ae7d40496793cf3d3e7890c9bbdc202263836b/falco_ubuntu_5.4.0-1044-azure_46.ko
  probe: output/39ae7d40496793cf3d3e7890c9bbdc202263836b/falco_ubuntu_5.4.0-1044-azure_46.o
```

Lowaiz: This will happen every time the extra version is being correctly parsed by the function `parseUbuntuExtraVersion` because you rebuild the extra version by splitting it with a dash as you parsed it.

EXONER4TED: Good catch! Let me work up a fix.

EXONER4TED: I went for just deduplicating the URL slice. In the worst case scenario, `parseUbuntuExtraVersion` just passes back to the user exactly what the user passed in since there is no error clause. In that case, I'd like to keep as many possible naming patterns and check for duplicates after the fact.

But it's probably also just good practice to deduplicate the slice anyway. If we can ensure that we're only getting 2 URLs to resolve, then we are also reducing the amount of `curl` calls and downloads/bandwidth.

EXONER4TED: There is an argument for just removing the bottom naming pattern altogether though. Since `parseUbuntuExtraVersion` has no error clause, it could be said that it will *always* parse the extra version correctly, and therefore always cause duplicates.

EXONER4TED: Yea, all good 😄 I know what you mean - and I think leaving it in there is fine in the event that the function is unable to split on `-` and just returns back whatever was passed in. I'm weary there may be some edge case in that scenario. And there should be no harm leaving it in with a `dedup` function cleaning it up.

cmd/root_options.go
```diff
@@ -20,7 +21,7 @@ type OutputOptions struct {
 type RootOptions struct {
 	Architecture     string   `validate:"required,oneof=amd64 arm64" name:"architecture"`
 	DriverVersion    string   `default:"master" validate:"eq=master|sha1|semver" name:"driver version"`
-	KernelVersion    uint16   `default:"1" validate:"omitempty,number" name:"kernel version"`
+	KernelVersion    string   `default:"1" validate:"required" name:"kernel version"`
```
dwindsor, line 24: Will changing this from `omitempty` to `required` break existing users that aren't supplying a value? Also, should we enforce `ascii` encoding as is done for `KernelRelease`?

test/aarch64/configs/ubuntu.yaml
```diff
@@ -1,5 +1,5 @@
 kernelversion: 199
-kernelrelease: 4.15.0-188
+kernelrelease: 4.15.0-188-generic
```
dwindsor, line 2: Hmm, not sure why we're having to change existing test input data like this? To be backwards compatible, I'd think that we'd want to be able to support this new format (with `-generic` in `kernelrelease`), and the old format (without `-generic`).

cmd/root_options.go
EXONER4TED: Good catch, let me set it back to `omitempty`. I had changed that very early on my development on the branch thinking I might have to go down a different path.

test/aarch64/configs/ubuntu.yaml
EXONER4TED: I can change back. Ideally, we won't see any without `-{flavor}`, but it is good to check for backwards compatibility.

pkg/driverbuilder/builder/ubuntu.go
```diff
@@ -296,69 +205,85 @@ func fetchUbuntuAWSKernelURLS(baseURL string, kr kernelrelease.KernelRelease, ke
...
-	ModuleDriverName     string
-	ModuleFullPath       string
-	BuildProbe           bool
-	BuildModule          bool
-	GCCVersion           string
+// parse the extraversion from the kernelrelease to retrieve the extraNumber and flavor
+// assume the flavor is "generic" if unable to parse the flavor
+// Example: Input -> "188-generic", Output -> "188", "generic"
+// NOTE: make sure the kernelrelease passed in appears *exactly* as `uname -r` output
+func parseUbuntuExtraVersion(extraversion string) (string, string) {
```
dwindsor, line 246: I think we should probably add a unit test for this, since these kernel versions will get pretty complex, as we've already seen.

Also, thanks for doing the leg work to get this function written! It was quite an adventure finding all the different flavors (I'm looking at you, `ubuntu-oracle`).

EXONER4TED: Will do, on it boss 👍

EXONER4TED: I wrote a starting point test file for the Ubuntu builder, was able to to achieve ~67% coverage. The tests aren't perfect, but they're a start.

cmd/root.go
FedeDP, line 131: Nitpick:
```suggestion
	flags.StringVar(&rootOpts.KernelVersion, "kernelversion", rootOpts.KernelVersion, "kernel version to build the module for, it's the string after the hash when you execute 'uname -v'")
```

pkg/driverbuilder/builder/ubuntu.go
```diff
@@ -16,98 +14,80 @@ import (
...
 var ubuntuTemplate string
 
-// TargetTypeUbuntuGeneric identifies the UbuntuGeneric target.
-const TargetTypeUbuntuGeneric Type = "ubuntu-generic"
+// TargetTypeUbuntu identifies the Ubuntu target.
+const TargetTypeUbuntu Type = "ubuntu"
 
-// TargetTypeUbuntuAWS identifies the UbuntuAWS target.
+// backwards compatibility
+const TargetTypeUbuntuGeneric Type = "ubuntu-generic"
```
FedeDP, line 21: If you are willing to open 2 more PRs:
* one in kernel-crawler to set `ubuntu` as target for ubuntu configs
* one in test-infra to replace all `ubuntu-{generic,aws}` with just `ubuntu` target

i think we can go as far as fully dropping ubuntu-generic and ubuntu-aws. What do you think?
We can:
* drop ubuntu-generic and ubuntu-aws support here
* open the 2 PRs and leave them open
* as soon as the new driverkit is released, we bump the driverkit version inside the test-infra PR
* then we can merge both the kernel-crawler and the test-infra one

WDYT?

cmd/root_options.go
```diff
@@ -128,12 +129,12 @@ func RootOptionsLevelValidation(level validator.StructLevel) {
 		level.ReportError(opts.KernelConfigData, "kernelConfigData", "KernelConfigData", "required_kernelconfigdata_with_target_vanilla", "")
 	}
 
-	if opts.KernelVersion == 0 && (opts.Target == builder.TargetTypeUbuntuAWS.String() || opts.Target == builder.TargetTypeUbuntuGeneric.String()) {
+	if opts.KernelVersion == "" && opts.Target == builder.TargetTypeUbuntu.String() {
```
FedeDP, line 132: In case we decide to keep ubuntu-generic and ubuntu-aws support, shouldn't they be listed here too?

FedeDP:
Also, i know it is getting annoying (sorry!) , but can you have a look at https://github.com/falcosecurity/test-infra/pull/766 ?

pkg/driverbuilder/builder/ubuntu.go
FedeDP: IMHO we can also keep backward compatibility for now, and break it later (like for driverkit 1.0.0); but the kernel-crawler patch would still be needed.

FedeDP:
Also, @EXONER4TED , i forgot to ask: did you try multiple kernel-crawler crawled ubuntu configs? (fixing of course the target).
I fear we might break something :/

cmd/root_options.go
EXONER4TED: Oh good catch, let me add that back.

EXONER4TED:
I can use more crawler configs if we'd like! I just did a:
```
python3 __init__.py crawl --distro Ubuntu --out_fmt driverkit
```
^ and that's been my input test on my local machine. I can fire the test off locally again on one of those lists you posted though.

FedeDP:
> I just did a:
>
> `python3 __init__.py crawl --distro Ubuntu --out_fmt driverkit`

This is exactly what i needed :) 
But you did have to change various `ubuntu-$flavour` with just `ubuntu` right?

pkg/driverbuilder/builder/ubuntu.go
EXONER4TED: Yea, I agree with you, and plan to open those PRs if/when this goes through 😄 

I would keep them for now, personally, for the sake of backwards compatibility like you said. But since Driverkit is not even to a 1.0 release, it's probably fine to quickly deprecate.

EXONER4TED:
Yea! I think it's a small/simple change to kernel-crawler 😄

FedeDP:
I will give it another look in ~1hr, but i think this LGTM now!

EXONER4TED:
Sounds good to me! The more 👀 the better. It's been very cool to get lots of feedback!

I can also open a PR for deprecating `ubuntu-aws` and `ubuntu-generic` which can just sit and be merged at leisure when ready.

pkg/driverbuilder/builder/ubuntu.go
```diff
@@ -16,98 +14,80 @@ import (
...
+	_, flavor := parseUbuntuExtraVersion(kr.Extraversion)
 
-	var urls []string
-	if c.KernelUrls == nil {
-		urls, err = ubuntuAWSHeadersURLFromRelease(kr, c.KernelVersion)
+	// handle hwe kernels, which resolve to "generic" urls under /linux-hwe
+	// Example: http://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux-hwe/linux-headers-4.18.0-24-generic_4.18.0-24.25~18.04.1_amd64.deb
+	headersPattern := ""
+	if flavor == "hwe" {
+		headersPattern = "linux-headers*generic"
```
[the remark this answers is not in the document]
EXONER4TED: I actually had done that at first. The problem is the `hwe` flavor is needed in the URL there. If you look, the baseURL is `http://mirrors.edge.kernel.org/ubuntu/pool/main/l/`, and the subdir is `linux-hwe`. So the flavor returned in `fetchUbuntuKernelURL` needs to be `hwe` for the url subdir to be correct.

But for some reason ubuntu stores the package under `/linux-hwe` and names the actual headers package `linux-headers*generic` lol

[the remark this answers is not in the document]
FedeDP: I am starting to hate on ubuntu :angry: ahah
That's ok for me then!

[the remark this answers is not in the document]
FedeDP: No problem, i think it's ok how it is now :)

FedeDP:
My approval is weak :/ @dwindsor

FedeDP (approved):
/approve

dwindsor (approved):

PR #172, opened 2022-06-23
Fix CI failure

dwindsor:
Signed-off-by: David Windsor <dwindsor@secureworks.com>

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area cmd

> /area pkg

> /area docs

> /area tests

#171 got approved lightning quick (before CI had finished), and integration_tests is failing. This fixes it.

Fixes #

```release-note

```

### DIFF
--- a/pkg/driverbuilder/builder/ubuntu.go
+++ b/pkg/driverbuilder/builder/ubuntu.go
@@ -416,7 +416,7 @@ func ubuntuGCCVersionFromKernelRelease(kr kernelrelease.KernelRelease) string {
 		}
 		return "6"
 	case "5":
-		if kr.PatchLevel >= "13" {
+		if kr.PatchLevel == "13" {
 			return "10"
 		}
 	}


dwindsor:
@leodido

FedeDP:
`gcc-10` is not available on our builder image: https://github.com/falcosecurity/driverkit#3-customize-gcc-version; neither it can be easily added because of glibc version.
This is something we should address using multiple builder images IMO.
See master failing integration tests (https://app.circleci.com/pipelines/github/falcosecurity/driverkit/304/workflows/503f6b45-2597-41d3-ac7d-107651cba53b/jobs/455/parallel-runs/0/steps/0-106)
```
ln -sf /usr/bin/gcc-10 /usr/bin/gcc 
cd /tmp/driver  
...
make[1]: gcc: Command not found
```

FedeDP:
I see, the GCCVersion is useful when using another builder image!
That makes sense! It would break without using a new builder image but driverkit would fail to build the driver in any case using the old GCC8.
:+1: from me!

dwindsor:
> I see, the GCCVersion is useful when using another builder image! That makes sense! It would break without using a new builder image but driverkit would fail to build the driver in any case using the old GCC8. 👍 from me!

I agree this is a hacky temporary fix, though. This problem should be addressed in the [base builder](https://github.com/falcosecurity/driverkit/blob/master/build/builder.Dockerfile), but adding gcc-10 there would require us to address underlying glibc/binutils compatibility issues - a significant change that deserves its own PR.

leodido (approved):

PR #173, opened 2022-06-24
chore(pkg/driverbuilder): port kernelrelease `version`, `patch` and `sublevel` to integers

FedeDP:
**What type of PR is this?**

/kind cleanup

/area pkg

Ported kernelrelease `version`, `patch` and `sublevel` to integers.
This should also fix driverkit master CI.

Also, improved internal API by passing to `Script` method the kernelrelease.

Fixes #

```release-note
NONE
```


pkg/driverbuilder/builder/amazonlinux.go
```diff
@@ -66,9 +66,6 @@ rm -Rf /tmp/kernel
 mkdir -p /tmp/kernel
 mv usr/src/kernels/*/* /tmp/kernel
 
-# Change current gcc
-ln -sf /usr/bin/gcc-{{ .GCCVersion }} /usr/bin/gcc
```
FedeDP, line 70: Dropped the `amazonGccVersionFromKernelRelease` function because it was just returning default GCC version.

pkg/driverbuilder/builder/centos.go
```diff
@@ -30,15 +30,15 @@ func (c centos) Script(cfg Config) (string, error) {
...
-    if cfg.KernelUrls == nil {
-        // Check (and filter) existing kernels before continuing
-        urls, err = getResolvingURLs(fetchCentosKernelURLS(kr))
-    } else {
-        urls, err = getResolvingURLs(cfg.KernelUrls)
-    }
-    if err != nil {
-        return "", err
-    }
+	if cfg.KernelUrls == nil {
```
FedeDP, line 33: :sob: saving this file has this side effect :)

pkg/driverbuilder/builder/debian.go
```diff
@@ -36,20 +36,24 @@ func (v debian) Script(c Config) (string, error) {
...
+		urls, err = getResolvingURLs(c.KernelUrls)
+	}
+	if err != nil {
+		return "", err
+	}
+	// We need:
+	// kernel devel
+	// kernel devel common
+	// kbuild package
+	if len(urls) < 3 {
```
FedeDP, line 56: We already check that all of these 3 are fetched, see:
* kernel-headers: https://github.com/falcosecurity/driverkit/blob/master/pkg/driverbuilder/builder/debian.go#L213
* kernel-common: https://github.com/falcosecurity/driverkit/blob/master/pkg/driverbuilder/builder/debian.go#L228
* https://github.com/falcosecurity/driverkit/blob/master/pkg/driverbuilder/builder/debian.go#L259
The latter check if at least 2 string submatch can be found, than uses `match[1]`.

pkg/driverbuilder/builder/debian.go
```diff
@@ -174,11 +178,11 @@ func debianHeadersURLFromRelease(kr kernelrelease.KernelRelease) ([]string, erro
 func fetchDebianHeadersURLFromRelease(baseURL string, kr kernelrelease.KernelRelease) ([]string, error) {
 	extraVersionPartial := strings.TrimSuffix(kr.FullExtraversion, "-"+kr.Architecture.String())
 	matchExtraGroup := kr.Architecture.String()
-	rmatch := `href="(linux-headers-%s\.%s\.%s%s-(%s)_.*(%s|all)\.deb)"`
+	rmatch := `href="(linux-headers-%d\.%d\.%d%s-(%s)_.*(%s|all)\.deb)"`
```
FedeDP, line 181: Please double check all of this; i should've fixed everything but a careful look is needed!

pkg/driverbuilder/builder/amazonlinux.go
FedeDP, line 70: Dropped `amazonGccVersionFromKernelRelease` function that was just using default gcc version.

pkg/driverbuilder/builder/ubuntu.go
```diff
@@ -410,13 +406,13 @@ ls -l probe.o
...
-		if kr.PatchLevel == "13" || kr.PatchLevel == "2" {
+	case 3:
+		if kr.PatchLevel == 13 || kr.PatchLevel == 2 {
 			return "4.8"
 		}
 		return "6"
-	case "5":
-		if kr.PatchLevel >= "13" {
+	case 5:
+		if kr.PatchLevel >= 13 {
```
FedeDP, line 415: This should fix current master.

FedeDP:
Given that both this one and #173  breaks newly developed builders, we can merge them before and i'll take care of rebasing these ones.

dwindsor:
> /cc @dwindsor :)

Thanks so much for your help on this!

pkg/driverbuilder/builder/build.go
```diff
@@ -15,3 +17,9 @@ type Build struct {
 	CustomBuilderImage string
 	KernelUrls         []string
 }
+
+func (b *Build) KernelReleaseFromBuildConfig() kernelrelease.KernelRelease {
+	kv := kernelrelease.FromString(b.KernelRelease)
```
dwindsor, line 22: Hmm, just curious: why does `kernelrelease.FromString` not accept an object of type `*Build` and then try to populate as many fields as it can (including, in this case, `.Architecture`)?

dwindsor, line 22: Hmm, just curious: why does `kernelrelease.FromString` not accept an object of type `*Build` and then try to populate as many fields as it can (including, in this case, `.Architecture`)?

dwindsor (approved):

PR #186, opened 2022-07-14
wip: chore: use docker manifest native for native docker images builds

FedeDP:
**What type of PR is this?**

/kind cleanup

/area build

This PR ports away from `docker buildx` all the docker images build and push CI to use `docker manifest` (https://docs.docker.com/engine/reference/commandline/manifest/) instead.

This allows us to highly speed up docker images builds because they're now built on native runners, instead of qemu emulated ones.
It is a somewhat pre-requisites for **multiple builder images** because current `images` job takes ~10mins with just a single builder image; imagine having like 5 of them...

As always with images build/push CI, i am actually not able to test it :/ 
As a workaround, i added a second commit to test build and push phase even from the PR; this will cause no breakage because images will be pushed with tags:
* git branch (ie: chore/use_docker_manifest_native_builds)
* git commit hash
ie: nobody will ever use them.

If CI works fine, i will rebase and remove the second commit.

Fixes #

```release-note
NONE
```


FedeDP:
I am trying to find ways to test this on my own fork :) 
Will remove the "wip" when this is ready.

FedeDP:
I was able to test this: https://app.circleci.com/pipelines/github/FedeDP/driverkit/160/workflows/6d880617-642f-4104-80be-4186352bfac1
It works now!
Here: https://hub.docker.com/r/fededp/driverkit/tags you can find tags for driverkit image, and here: https://hub.docker.com/r/fededp/driverkit-builder for builder image; as you can see, manifest is working and we got single arch images + final multi arch images.

CI time splits before:
* total: ~40min
* build-test: ~25m
* build-test-arm64: ~6m
* images: ~13m

CI time splits after:
* total: ~30m
* build-test: ~25m
* build-test-arm64: ~6m
* build-images: 3m
* build-images-arm64: 2.5m
*  images: 0.5m

Note: total is not the sum of other steps, because many steps run in parallel.
Only pain point now is the build-test job that takes ages because of its internal qemu cross builds tests.
If we need to free up more minutes, we can think about skipping qemu cross builds; for now, i'd leave them enabled.

FedeDP:
I am now testing the `docker-layer-caching` circleci option: https://circleci.com/docs/docker-layer-caching

dwindsor (approved):

PR #193, opened 2022-07-27
chore: improve tests and readme

FedeDP:
**What type of PR is this?**

/kind cleanup

/area cmd
/area docs

The PR enhances test suite for cmd package, leveraging text.template golang package (the same we use for builder's template scripts!).

Basically, test cases output is now a template and is run against a dynamically loaded template data.
This means:
* no more need to add any new target to `cmd/testdata/completion-targets.txt`
* no more need to add any possible new architecture to `cmd/testdata` files
* no more need to add any added/removed cmd line flags to `cmd/testdata` files

Just updating a single file in `cmd/testdata/templates` will be enough!

Moreover, updated README splitting examples and builder info to different files.

Finally, actually print list of supported architectures and targets in driverkit help text.

Fixes #

```release-note
NONE
```


Example_configs.md
```diff
@@ -0,0 +1,289 @@
...
+```
+
+## ubuntu-aws
+
+Example configuration file to build both the Kernel module and eBPF probe for Ubuntu AWS.
+
+```yaml
+kernelrelease: 4.15.0-1057-aws
+kernelversion: 59
+target: ubuntu-aws
```
Lowaiz, line 36:
```suggestion
target: ubuntu
```
Same as above.

Example_configs.md
```diff
@@ -0,0 +1,289 @@
...
+driverversion: master
+```
+
+## ubuntu-generic
+Example configuration file to build both the Kernel module and eBPF probe for Ubuntu generic.
+
+```yaml
+kernelrelease: 4.15.0-72-generic
+kernelversion: 81
+target: ubuntu-generic
```
Lowaiz, line 22:
```suggestion
target: ubuntu
```
As the target `ubuntu-generic` is the previous format, maybe we should doc only the new one ?

Lowaiz: We should, at least, indicate that this is the old format.

Example_configs.md
Lowaiz, line 36: Same as above.
```suggestion
target: ubuntu
```

Example_configs.md
FedeDP: Agree, perhaps a note can be helpful there. We can remove these altogether too IMO!
@dwindsor what do you think?

Example_configs.md
```diff
@@ -0,0 +1,289 @@
+# Example Configs
```
FedeDP, line 1: This is just moved out from main README.

cmd/cli_test.go
```diff
@@ -345,6 +333,81 @@ func run(t *testing.T, test testCase) {
 	}
 }
 
+// Simple explanation:
```
FedeDP, line 336: CLI test will now use a text.template based approach: expected output txt files will have access to template data in the same manner as, for example, our builder's template scripts.

This allow us to execute templates against dynamically loaded data, like list of supported architectures, list of supported targets, current architecture...

@dwindsor i think the same approach could also be used for kernelurls formatting, if that is still a desired feature :)

cmd/root.go
```diff
@@ -37,34 +37,34 @@ func persistentValidateFunc(rootCommand *RootCmd, rootOpts *RootOptions) func(c
...
-                            value = viper.GetString(nestedName)
-                        }
-                    }
-                    // set the value, if any, otherwise let the default
-                    if value != "" {
-                        rootCommand.c.Flags().Set(name, value)
-                    }
-                }
-            }
+			if name := f.Name; !skip[name] {
```
FedeDP, line 40: Fixed indentation... :/ goland automatically did this.

cmd/root.go
```diff
@@ -126,11 +129,11 @@ func NewRootCmd() *RootCmd {
 
 	flags.StringVar(&rootOpts.Output.Module, "output-module", rootOpts.Output.Module, "filepath where to save the resulting kernel module")
 	flags.StringVar(&rootOpts.Output.Probe, "output-probe", rootOpts.Output.Probe, "filepath where to save the resulting eBPF probe")
-	flags.StringVar(&rootOpts.Architecture, "architecture", runtime.GOARCH, "target architecture for the built driver")
+	flags.StringVar(&rootOpts.Architecture, "architecture", runtime.GOARCH, "target architecture for the built driver, one of ["+strings.Join(builder.SupportedArchs, ",")+"]")
```
FedeDP, line 132: Print available values.

cmd/root.go
```diff
@@ -126,11 +129,11 @@ func NewRootCmd() *RootCmd {
...
 	flags.StringVar(&rootOpts.DriverVersion, "driverversion", rootOpts.DriverVersion, "driver version as a git commit hash or as a git tag")
 	flags.StringVar(&rootOpts.KernelVersion, "kernelversion", rootOpts.KernelVersion, "kernel version to build the module for, it's the numeric value after the hash when you execute 'uname -v'")
 	flags.StringVar(&rootOpts.KernelRelease, "kernelrelease", rootOpts.KernelRelease, "kernel release to build the module for, it can be found by executing 'uname -v'")
-	flags.StringVarP(&rootOpts.Target, "target", "t", rootOpts.Target, "the system to target the build for")
+	flags.StringVarP(&rootOpts.Target, "target", "t", rootOpts.Target, "the system to target the build for, one of ["+strings.Join(targets, ",")+"]")
```
FedeDP, line 136: Print available values.

cmd/root.go
```diff
@@ -142,10 +145,11 @@ func NewRootCmd() *RootCmd {
 	// Flag annotations and custom completions
 	rootCmd.MarkFlagFilename("config", viper.SupportedExts...)
 	rootCmd.RegisterFlagCompletionFunc("target", func(c *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-		targets := builder.BuilderByTarget.Targets()
-		sort.Strings(targets)
 		return targets, cobra.ShellCompDirectiveDefault
 	})
+	rootCmd.RegisterFlagCompletionFunc("architecture", func(c *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
FedeDP, line 150: Added completion function for `architecture`.

cmd/testdata/autohelp.txt
```diff
@@ -1,35 +1,11 @@
...
-  -l, --loglevel string           log level (default "info")
-      --moduledevicename string   kernel module device name (the default is falco, so the device will be under /dev/falco*) (default "falco")
-      --moduledrivername string   kernel module driver name, i.e. the name you see when you check installed modules via lsmod (default "falco")
-      --output-module string      filepath where to save the resulting kernel module
-      --output-probe string       filepath where to save the resulting eBPF probe
-      --proxy string              the proxy to use to download data
-  -t, --target string             the system to target the build for
-      --timeout int               timeout in seconds (default 120)
+{{ .Flags }}
   -v, --version                   version for driverkit
```
FedeDP, line 9: This is kept out of `{{ .Flags }}` because sometimes it is missing.

pkg/driverbuilder/builder/README.md
```diff
@@ -0,0 +1,153 @@
+# Builders
```
FedeDP, line 1: Split out from main README.

pkg/kernelrelease/kernelrelease.go
```diff
@@ -1,25 +1,53 @@
...
 
 var (
 	kernelVersionPattern = regexp.MustCompile(`(?P<fullversion>^(?P<version>0|[1-9]\d*)\.(?P<patchlevel>0|[1-9]\d*)\.(?P<sublevel>0|[1-9]\d*))(?P<fullextraversion>-(?P<extraversion>0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-_]*))*)?(\+[0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*)?$`)
 )
 
+// Architectures is a Map [Architecture] -> non-deb-ArchitectureString
+type Architectures map[Architecture]string
+
+// SupportedArchs enforces the duality of architecture->non-deb one when adding a new one
+var SupportedArchs = Architectures{
```
FedeDP, line 19: The idea is that adding support for a new architecture is as easy as adding a new entry here.
Note: i made practically impossible to forget to add also the Non-deb name (previously handled in a `ToNonDeb()` method switch case statement).

Example_configs.md
dwindsor: I think we should probably start deprecating the old `target` formatting before our 1.0 release, so a note would be good! @FedeDP do you mean remove this file altogether?

FedeDP: Uh no, i meant remove the lines about `ubuntu-aws` and `ubuntu-generic` :)

cmd/cli_test.go
dwindsor: I really like the approach of using `go:embed` in the builders, so yes this is a good change imo!

@FedeDP with respect to `--kernelurls` formatting, what are your thoughts on deprecating all crawling functionality from driverkit altogether? IOW, force folks to use kernel-crawler ;). Only reason I suggest that is, headers package formatting is a hard problem and I feel like it's more maintainable long-term for the project to only have this logic in one place? But, I'm not sure, what are your thoughts? Would this annoy existing users you think?

FedeDP: > I really like the approach of using go:embed in the builders, so yes this is a good change imo!

I am using `text.template` here though :D 
> what are your thoughts on deprecating all crawling functionality from driverkit altogether?

Yeah, i think that would be shocking for existing users; i think some kind of basic headers discovery logic should remain nested in driverkit for a while, perhaps we can add a "deprecation" notice in the README?

dwindsor: Yeah I was thinking longer term maybe, I'm just thinking about the explosion of distros/builders we're going to (hopefully!) be getting, and the burden of maintaining them in multiple places (new flavors added, vendor repo reorgs, etc). At this point it's not an issue at all, just felt like it was worth calling out.

FedeDP: Fully agree with you :/ 
I think that we should properly state that intenral logic is a fallback one, but the preferred method is passing `kernelurls` parameter (be it manually found or taken from kernel-crawler!)

cmd/cli_test.go
```diff
@@ -345,6 +334,81 @@ func run(t *testing.T, test testCase) {
...
+// basically we leverage text.template to execute templates for test data.
+// testTemplateData is the actual template structure used to fill the test requested output.
+// It internally stores all the help text, split in section { Desc, Usage, Commands, Flags, Info }
+// plus the flagsTemplateData (that is the template data used to fill Flags fields basically).
+type testTemplateData struct {
+	Desc     string
+	Usage    string
+	Commands string
+	Flags    string
```
FedeDP, line 346: Note that flags is itself a templated script ;) therefore we need both the
* `testTemplateData` -> to be passed to actual test template
* `flagsTemplateData` -> to be used by testTemplateData to compute its `Flags` field.

Example_configs.md
FedeDP: Added the notes!

cmd/cli_test.go
FedeDP: I am going to add a NOTE about that!

FedeDP: Done!

dwindsor (approved):

PR #197, opened 2022-08-16
feat(kebernetes-incluster): Adding kubernetes in cluster subcommand

Lowaiz:
**What type of PR is this?**

/kind feature

/area cmd

/area pkg

* Adding a sub-command to use Kubernetes feature inside a kube pods with the appropriate service token/account.
* Adding configs/flags options to handle non root execution and pull secrets when using custom builder image.
* Fixing Kubenertes execution and probe module extraction. (Now extract module and probe properly, than delete the configmap and the pod)

~~To make non root execution works, I had to make a sub docker image builder with this owner change:~~
```
FROM falcosecurity/driverkit-builder

RUN chown -R nobody /usr/bin
```
~~I have to do this because driverkit's generated build script run a `ln -sf /usr/bin/gcc ...`, requiring root right/access.~~
~~Can you see another workaround ? Except to rewrite the builder Dockerfile and scripts to make them non-root executable.~~

None

```release-note
* New sub-command `kubernetes-in-cluster` or `k8s-ic`
* New Kubernetes flags: 
  * namespace: Namespace where to execute the builder pod
  * run-as-user: User Id of the pod
  * imagePullSecret: Pull secret for custom builder image
```

### DIFF
--- a/pkg/driverbuilder/builder/templates/archlinux.sh
+++ b/pkg/driverbuilder/builder/templates/archlinux.sh
@@ -21,13 +21,10 @@ rm -Rf /tmp/kernel
 mkdir -p /tmp/kernel
 mv usr/lib/modules/*/build/* /tmp/kernel
 
-# Change current gcc
-ln -sf /usr/bin/gcc-{{ .GCCVersion }} /usr/bin/gcc
-
 {{ if .BuildModule }}
 # Build the module
 cd {{ .DriverBuildDir }}
-make KERNELDIR=/tmp/kernel
+make CC=/usr/bin/gcc-{{ .GCCVersion }} KERNELDIR=/tmp/kernel
 mv {{ .ModuleDriverName }}.ko {{ .ModuleFullPath }}
 strip -g {{ .ModuleFullPath }}
 # Print results
@@ -37,6 +34,6 @@ modinfo {{ .ModuleFullPath }}
 {{ if .BuildProbe }}
 # Build the eBPF probe
 cd {{ .DriverBuildDir }}/bpf
-make LLC=/usr/bin/llc-7 CLANG=/usr/bin/clang-7 CC=/usr/bin/gcc KERNELDIR=/tmp/kernel
+make LLC=/usr/bin/llc-7 CLANG=/usr/bin/clang-7 CC=/usr/bin/gcc-{{ .GCCVersion }} KERNELDIR=/tmp/kernel
 ls -l probe.o
 {{ end }}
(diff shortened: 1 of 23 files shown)


FedeDP:
Hi @Lowaiz ! Thanks for this great PR!

> I have to do this because driverkit's generated build script run a ln -sf /usr/bin/gcc ..., requiring root right/access.
Can you see another workaround ? Except to rewrite the builder Dockerfile and scripts to make them non-root executable.

Mmmh perhaps we could override CC env variable?
For reference, the ubuntu one would become:
```
{{ if .BuildModule }}
# Build the module
cd {{ .DriverBuildDir }}
make CC=/usr/bin/gcc-{{ .GCCVersion }} KERNELDIR=$sourcedir
mv {{ .ModuleDriverName }}.ko {{ .ModuleFullPath }}
strip -g {{ .ModuleFullPath }}
# Print results
modinfo {{ .ModuleFullPath }}
{{ end }}
```

This is what we already do for clang btw, eg (taken from same ubuntu script):
```
{{ if .BuildProbe }}
# Build the eBPF probe
cd {{ .DriverBuildDir }}/bpf
if [[ -x /usr/bin/llc ]]; then
	LLC_BIN=/usr/bin/llc
else
	LLC_BIN=/usr/bin/llc-7
fi

if [[ -x /usr/bin/clang ]]; then
	CLANG_BIN=/usr/bin/clang
else
	CLANG_BIN=/usr/bin/clang-7
fi

make LLC=$LLC_BIN CLANG=$CLANG_BIN CC=/usr/bin/gcc-8 KERNELDIR=$sourcedir
```

Lowaiz:
Hi @FedeDP,

> I think this would be a valuable improvement that perhaps could go in a different PR, wdyt?

This enhancement make usable the non-root security context, so I guess we should put them together. 

I already pushed the changes, but I can move them elsewhere if needed

FedeDP:
> This enhancement make usable the non-root security context, so I guess we should put them together.

Yep, no problem on my side :)

pkg/driverbuilder/builder/templates/archlinux.sh
```diff
@@ -37,6 +34,6 @@ modinfo {{ .ModuleFullPath }}
 {{ if .BuildProbe }}
 # Build the eBPF probe
 cd {{ .DriverBuildDir }}/bpf
-make LLC=/usr/bin/llc-7 CLANG=/usr/bin/clang-7 CC=/usr/bin/gcc KERNELDIR=/tmp/kernel
+make LLC=/usr/bin/llc-7 CLANG=/usr/bin/clang-7 CC=/usr/bin/gcc-{{ .GCCVersion }} KERNELDIR=/tmp/kernel
```
FedeDP, line 37: I think we can also cleanup `CC=` from the various bpf build scripts, as it is useless!

Lowaiz: Applied everywhere :heavy_check_mark:

FedeDP: That's a super quick response time :rocket:

FedeDP (approved):
/approve

PR #201, opened 2022-08-26
chore: support any ubuntu flavor as a target

FedeDP:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area build

/area cmd

/area pkg

> /area docs

> /area tests

**What this PR does / why we need it**:

This PR is part of a wider effort to support any ubuntu flavor to build Falco drivers; see:
* https://github.com/falcosecurity/falco/pull/2178
* https://github.com/falcosecurity/kernel-crawler/pull/38

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
new(cmd): support any ubuntu flavor to be used as a target by internally managing it as just ubuntu
```

PR #202, opened 2022-08-26
new: multiple builders

### LINKED ISSUE #194: Multiple builder images
**Motivation**

As most issues report, like #145  and #126, we are not able to build most recent kernels because our builder image ships with gcc8 as most recent gcc version.

**Feature**

We should be able to properly support multiple builder images, and have an automatic logic to choose the correct one.
The logic should be as simple as possible, matching:
* the requested GCC version (just like we currently have like: `UbuntuGCCVersionFromKernelRelease`
* the requested Clang version (again, we currently have something similar)
* if none of the above is specified, a simple logic based on the requested kernelrelease

While not perfect, this solution is simple, involves little code, and should be quite effective.

**Alternatives**

Another idea could be to host builder images in a different repository, pushed to dockerhub, and avoid having any "automatic" builder image; instead, always relying on the user to provide a config with the correct builder image.
In this sense, it would be a `kernel-crawler` duty to provide in its output json, the correct builder image for each config.
Basically, kernel-crawler should iterate a call to driverkit using each supported builder image, until it successfully builds the kernel, for each crawled kernel.

While tedious, this has the enormous advantage that the "builder image discovery" is made only once, at the kernel-crawler level.
And driverkit will always expect the user to provide a builderimage, making the param mandatory.

This solution is more complex, but has the benefit of (in theory) being able to always discover the right builder image and to generate driverkit configs with already the builder image set.
The downside is:
* much more complex
* involves code somewhere else (not a real downside imo)
* greatly increase kernel-crawler job duration -> i'd expect around some days of work, because we basically run N builds in the worst case (where N is number of builder images) for each crawled kernel...

/cc @dwindsor @LucaGuerra @EXONER4TED @leodido

FedeDP:
**What type of PR is this?**

/kind feature

/area build

/area cmd

/area pkg

/area docs

> /area tests

This PR adds support for multiple builder images.
At the moment, 3 of them are supported:
* the old `buster` one
* a `bullseye` one
* a `bookworm` one

In my tests, they allowed to build basically every 5.X kernel for ubuntu, using config stolen by test-infra, with kernelurls (aside from a bunch of `kernel headers not found` that probably mean that Ubuntu dropped support for these kernels).

More tests can be useful of course :) 

I will try to auto-review the PR to make it easier to follow the code.

Moreover, a new option `gccversion` is now available, to force a gccversion (and therefore a builder image) for a given build.

Note that i completely dropped support for multiple clang versions, removing clang from our support matrix; if an eBPF probe cannot be built with clang shipped in any of the supported builder distros, then it is a bug of libs and an issue should be opened.  
We support clang7+ for our eBPF probe in libs.

Fixes #194

Most probably the CI will fail because the needed builder images are not present (given that they will be pushed only once the PR is merged :D )
We might want to avoid running integration-tests on this one, to avoid the errors; i don't really know.

```release-note
new: support multiple builder images
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,8 +17,8 @@ jobs:
           name: Test
           command: make test
       - run:
-          name: Integration tests # test cross build too!
-          command: make -e TARGET_TEST_ARCH=* integration_test
+          name: Integration tests
+          command: make integration_test
       - run:
           name: Prepare Artifacts
           command: |
--- a/Makefile
+++ b/Makefile
@@ -115,6 +115,7 @@ manifest/latest:
 
 .PHONY: test
 test:
+	go clean -testcache
 	go test -v -cover -race ./...
 	go test -v -cover -buildmode=pie ./cmd
 
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ ifeq ($(COMMITS_FROM_GIT_TAG),0)
 	endif
 endif
 
-DOCKER_ORG ?= falcosecurity
+DOCKER_ORG ?= fededp
 
 BUILDERS := $(patsubst build/builder_%.Dockerfile,%,$(wildcard build/builder_*.Dockerfile))
 
@@ -32,7 +32,7 @@ IMAGE_NAME_DRIVERKIT_REF := $(IMAGE_NAME_DRIVERKIT):$(GIT_REF)_$(shell uname -m)
 IMAGE_NAME_DRIVERKIT_COMMIT := $(IMAGE_NAME_DRIVERKIT):$(GIT_COMMIT)_$(shell uname -m)
 IMAGE_NAME_DRIVERKIT_LATEST := $(IMAGE_NAME_DRIVERKIT):latest_$(shell uname -m)
 
-LDFLAGS := -X github.com/falcosecurity/driverkit/pkg/version.buildTime=$(shell date +%s) -X github.com/falcosecurity/driverkit/pkg/version.gitCommit=${GIT_COMMIT} -X github.com/falcosecurity/driverkit/pkg/version.gitTag=$(if ${GIT_TAG},${GIT_TAG},v0.0.0) -X github.com/falcosecurity/driverkit/pkg/version.commitsFromGitTag=${COMMITS_FROM_GIT_TAG} -X github.com/falcosecurity/driverkit/pkg/builder.BaseImage=${IMAGE_NAME_BUILDER_BASE}:$(GIT_COMMIT)
+LDFLAGS := -X github.com/falcosecurity/driverkit/pkg/version.buildTime=$(shell date +%s) -X github.com/falcosecurity/driverkit/pkg/version.gitCommit=${GIT_COMMIT} -X github.com/falcosecurity/driverkit/pkg/version.gitTag=$(if ${GIT_TAG},${GIT_TAG},v0.0.0) -X github.com/falcosecurity/driverkit/pkg/version.commitsFromGitTag=${COMMITS_FROM_GIT_TAG} -X github.com/falcosecurity/driverkit/pkg/driverbuilder/builder.BaseImage=${IMAGE_NAME_BUILDER_BASE}:$(GIT_COMMIT)
 
 TARGET_TEST_ARCH ?= $(shell uname -m)
 test_configs := $(wildcard test/$(TARGET_TEST_ARCH)/configs/*.yaml)
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ ifeq ($(COMMITS_FROM_GIT_TAG),0)
 	endif
 endif
 
-DOCKER_ORG ?= fededp
+DOCKER_ORG ?= falcosecurity
 
 BUILDERS := $(patsubst build/builder_%.Dockerfile,%,$(wildcard build/builder_*.Dockerfile))
 
--- a/Makefile
+++ b/Makefile
@@ -18,11 +18,13 @@ endif
 
 DOCKER_ORG ?= falcosecurity
 
-IMAGE_NAME_BUILDER ?= docker.io/$(DOCKER_ORG)/driverkit-builder
+BUILDERS := $(patsubst build/builder_%.Dockerfile,%,$(wildcard build/builder_*.Dockerfile))
 
-IMAGE_NAME_BUILDER_REF := $(IMAGE_NAME_BUILDER):$(GIT_REF)_$(shell uname -m)
-IMAGE_NAME_BUILDER_COMMIT := $(IMAGE_NAME_BUILDER):$(GIT_COMMIT)_$(shell uname -m)
-IMAGE_NAME_BUILDER_LATEST := $(IMAGE_NAME_BUILDER):latest_$(shell uname -m)
+IMAGE_NAME_BUILDER_BASE ?= docker.io/$(DOCKER_ORG)/driverkit-builder
+
+IMAGE_NAME_SUFFIX_REF := ":$(GIT_REF)_$(shell uname -m)"
+IMAGE_NAME_SUFFIX_COMMIT := ":$(GIT_COMMIT)_$(shell uname -m)"
+IMAGE_NAME_SUFFIX_LATEST := ":latest_$(shell uname -m)"
 
 IMAGE_NAME_DRIVERKIT ?= docker.io/$(DOCKER_ORG)/driverkit
 
@@ -30,7 +32,7 @@ IMAGE_NAME_DRIVERKIT_REF := $(IMAGE_NAME_DRIVERKIT):$(GIT_REF)_$(shell uname -m)
 IMAGE_NAME_DRIVERKIT_COMMIT := $(IMAGE_NAME_DRIVERKIT):$(GIT_COMMIT)_$(shell uname -m)
 IMAGE_NAME_DRIVERKIT_LATEST := $(IMAGE_NAME_DRIVERKIT):latest_$(shell uname -m)
 
-LDFLAGS := -X github.com/falcosecurity/driverkit/pkg/version.buildTime=$(shell date +%s) -X github.com/falcosecurity/driverkit/pkg/version.gitCommit=${GIT_COMMIT} -X github.com/falcosecurity/driverkit/pkg/version.gitTag=$(if ${GIT_TAG},${GIT_TAG},v0.0.0) -X github.com/falcosecurity/driverkit/pkg/version.commitsFromGitTag=${COMMITS_FROM_GIT_TAG} -X github.com/falcosecurity/driverkit/pkg/driverbuilder.BuilderBaseImage=$(IMAGE_NAME_BUILDER):$(GIT_COMMIT)
+LDFLAGS := -X github.com/falcosecurity/driverkit/pkg/version.buildTime=$(shell date +%s) -X github.com/falcosecurity/driverkit/pkg/version.gitCommit=${GIT_COMMIT} -X github.com/falcosecurity/driverkit/pkg/version.gitTag=$(if ${GIT_TAG},${GIT_TAG},v0.0.0) -X github.com/falcosecurity/driverkit/pkg/version.commitsFromGitTag=${COMMITS_FROM_GIT_TAG} -X github.com/falcosecurity/driverkit/pkg/builder.BaseImage=${IMAGE_NAME_BUILDER_BASE}:$(GIT_COMMIT)
 
 TARGET_TEST_ARCH ?= $(shell uname -m)
 test_configs := $(wildcard test/$(TARGET_TEST_ARCH)/configs/*.yaml)
@@ -57,7 +59,9 @@ image/all: image/builder image/driverkit
 
 .PHONY: image/builder
 image/builder:
-	$(DOCKER) buildx build -o type=image,push="false" -f build/builder.Dockerfile .
+	$(foreach b,$(BUILDERS),\
+		$(DOCKER) buildx build -o type=image,push="false" -f build/builder_$b.Dockerfile . ; \
+    )
 
 .PHONY: image/driverkit
 image/driverkit:
@@ -67,7 +71,9 @@ push/all: push/builder push/driverkit
 
 .PHONY: push/builder
 push/builder:
-	$(DOCKER) buildx build --push -t "$(IMAGE_NAME_BUILDER_REF)" -t "$(IMAGE_NAME_BUILDER_COMMIT)" -f build/builder.Dockerfile .
+	$(foreach b,$(BUILDERS),\
+		$(DOCKER) buildx build --push -t "$(IMAGE_NAME_BUILDER_BASE)_$b$(IMAGE_NAME_SUFFIX_REF)" -t "$(IMAGE_NAME_BUILDER_BASE)_$b$(IMAGE_NAME_SUFFIX_COMMIT)" -f build/builder_$b.Dockerfile . ; \
+	)
 
 .PHONY: push/driverkit
 push/driverkit:
@@ -75,17 +81,21 @@ push/driverkit:
 
 .PHONY: push/latest
 push/latest:
-	$(DOCKER) buildx build --push -t "$(IMAGE_NAME_BUILDER_LATEST)" -f build/builder.Dockerfile .
+	$(foreach b,$(BUILDERS),\
+		$(DOCKER) buildx build --push -t "$(IMAGE_NAME_BUILDER_BASE)_$b$(IMAGE_NAME_SUFFIX_LATEST)" -f build/builder_$b.Dockerfile . ; \
+	)
 	$(DOCKER) buildx build --push -t "$(IMAGE_NAME_DRIVERKIT_LATEST)" -f build/driverkit.Dockerfile .
 
 manifest/all: manifest/builder manifest/driverkit
 
 .PHONY: manifest/builder
 manifest/builder:
-	$(DOCKER) manifest create $(IMAGE_NAME_BUILDER):$(GIT_REF) $(IMAGE_NAME_BUILDER):$(GIT_REF)_x86_64 $(IMAGE_NAME_BUILDER):$(GIT_REF)_aarch64
-	$(DOCKER) manifest push $(IMAGE_NAME_BUILDER):$(GIT_REF)
-	$(DOCKER) manifest create $(IMAGE_NAME_BUILDER):$(GIT_COMMIT) $(IMAGE_NAME_BUILDER):$(GIT_COMMIT)_x86_64 $(IMAGE_NAME_BUILDER):$(GIT_COMMIT)_aarch64
-	$(DOCKER) manifest push $(IMAGE_NAME_BUILDER):$(GIT_COMMIT)
+	$(foreach b,$(BUILDERS),\
+		$(DOCKER) manifest create "$(IMAGE_NAME_BUILDER_BASE)_$b:$(GIT_REF)" $(IMAGE_NAME_BUILDER_BASE)_$b:$(GIT_REF)_x86_64 $(IMAGE_NAME_BUILDER_BASE)_$b:$(GIT_REF)_aarch64 ; \
+		$(DOCKER) manifest push "$(IMAGE_NAME_BUILDER_BASE)_$b:$(GIT_REF)" ; \
+		$(DOCKER) manifest create $(IMAGE_NAME_BUILDER_BASE)_$b:$(GIT_COMMIT) $(IMAGE_NAME_BUILDER_BASE)_$b:$(GIT_COMMIT)_x86_64 $(IMAGE_NAME_BUILDER_BASE)_$b:$(GIT_COMMIT)_aarch64 ; \
+		$(DOCKER) manifest push $(IMAGE_NAME_BUILDER_BASE)_$b:$(GIT_COMMIT) ; \
+	)
 
 .PHONY: manifest/driverkit
 manifest/driverkit:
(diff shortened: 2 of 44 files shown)


FedeDP:
Failing tests from `make test` will be fixed with #201; i will rebase then.

.circleci/config.yml
```diff
@@ -17,8 +17,8 @@ jobs:
           name: Test
           command: make test
       - run:
-          name: Integration tests # test cross build too!
-          command: make -e TARGET_TEST_ARCH=* integration_test
+          name: Integration tests
```
FedeDP, line 20: Dropped cross-qemu build during integration tests to free up CI space/time for new builder images.

Makefile
```diff
@@ -18,19 +18,21 @@ endif
 
 DOCKER_ORG ?= falcosecurity
 
-IMAGE_NAME_BUILDER ?= docker.io/$(DOCKER_ORG)/driverkit-builder
+BUILDERS := $(patsubst build/builder_%.Dockerfile,%,$(wildcard build/builder_*.Dockerfile))
```
FedeDP, line 21: List all builders name.
NOTE that builder image name should always be:
`builder_$osname.Dockerfile`

Makefile
```diff
@@ -18,19 +18,21 @@ endif
...
 
 IMAGE_NAME_DRIVERKIT_REF := $(IMAGE_NAME_DRIVERKIT):$(GIT_REF)_$(shell uname -m)
 IMAGE_NAME_DRIVERKIT_COMMIT := $(IMAGE_NAME_DRIVERKIT):$(GIT_COMMIT)_$(shell uname -m)
 IMAGE_NAME_DRIVERKIT_LATEST := $(IMAGE_NAME_DRIVERKIT):latest_$(shell uname -m)
 
-LDFLAGS := -X github.com/falcosecurity/driverkit/pkg/version.buildTime=$(shell date +%s) -X github.com/falcosecurity/driverkit/pkg/version.gitCommit=${GIT_COMMIT} -X github.com/falcosecurity/driverkit/pkg/version.gitTag=$(if ${GIT_TAG},${GIT_TAG},v0.0.0) -X github.com/falcosecurity/driverkit/pkg/version.commitsFromGitTag=${COMMITS_FROM_GIT_TAG} -X github.com/falcosecurity/driverkit/pkg/driverbuilder.BuilderBaseImage=${IMAGE_NAME_BUILDER_COMMIT}
+LDFLAGS := -X github.com/falcosecurity/driverkit/pkg/version.buildTime=$(shell date +%s) -X github.com/falcosecurity/driverkit/pkg/version.gitCommit=${GIT_COMMIT} -X github.com/falcosecurity/driverkit/pkg/version.gitTag=$(if ${GIT_TAG},${GIT_TAG},v0.0.0) -X github.com/falcosecurity/driverkit/pkg/version.commitsFromGitTag=${COMMITS_FROM_GIT_TAG} -X github.com/falcosecurity/driverkit/pkg/driverbuilder/builder.BaseImage=${IMAGE_NAME_BUILDER_BASE}:$(GIT_COMMIT)
```
FedeDP, line 35: Moved BaseImage from driverbuilder package to builder.

Makefile
```diff
@@ -111,7 +124,7 @@ integration_test: $(test_configs)
 
 .PHONY: $(test_configs)
 $(test_configs): ${driverkit}
-	${driverkit} docker -c $@ --builderimage falcosecurity/driverkit-builder:latest -l debug --timeout 600
+	${driverkit} docker -c $@ --builderimage auto:latest -l debug --timeout 600
```
FedeDP, line 127: `auto:$imageTag` is also a supported builder image name now: it basically means "use automatic image selection but force an imageTag".

build/builder_buster.Dockerfile
```diff
@@ -43,12 +43,6 @@ RUN apt-get update \
 
 RUN if [ "$TARGETARCH" = "amd64" ] ; then apt-get install -y --no-install-recommends libmpx2; fi
 
-# Install clang 12
```
FedeDP, line 46: Do not install clang-12 anymore. We will use clang-7 here.

docs/driverkit.md
```diff
@@ -6,29 +6,6 @@ A command line tool to build Falco kernel modules and eBPF probes.
 driverkit
 ```
 
-### Options
```
FedeDP, line 9: Cleanup.

docs/driverkit_docker.md
```diff
@@ -6,29 +6,6 @@ Build Falco kernel modules and eBPF probes against a docker daemon.
 driverkit docker [flags]
 ```
 
-### Options
```
FedeDP, line 9: Cleanup.

docs/driverkit_kubernetes-in-cluster.md
```diff
@@ -6,32 +6,6 @@ Build Falco kernel modules and eBPF probes against a Kubernetes cluster inside a
 driverkit kubernetes-in-cluster [flags]
 ```
 
-### Options
```
FedeDP, line 9: Cleanup.

docs/driverkit_kubernetes.md
```diff
@@ -6,48 +6,6 @@ Build Falco kernel modules and eBPF probes against a Kubernetes cluster.
 driverkit kubernetes [flags]
 ```
 
-### Options
```
FedeDP, line 9: Cleanup.

pkg/driverbuilder/builder/README.md
```diff
@@ -124,23 +123,48 @@ Indeed, the hardest part is fetching the kernel headers urls for each distro.
 
 ### 3. Customize GCC version
```
FedeDP, line 124: Updated doc!

pkg/driverbuilder/builder/README.md
```diff
@@ -28,16 +29,15 @@ touch pkg/driverbuilder/builder/archlinux.go
...
+more than one target at time. For example, the minikube builder is just a vanilla one.
 
 Once you have the constant, you will need to add it to the `BuilderByTarget` map.
 
-
 Open your file and you will need to have something like this:
 
 ```go
 // TargetTypeArchLinux identifies the Arch Linux target.
+/// NOTE: the target name should exactly match the /etc/os-release ID value.
```
FedeDP, line 40: Mentions /etc/os-release :)

pkg/driverbuilder/builder/amazonlinux.go
```diff
@@ -314,14 +312,3 @@ func bunzip(data io.Reader) (res []byte, err error) {
 
 	return
 }
-
-func amazonLLVMVersionFromKernelRelease(kr kernelrelease.KernelRelease) string {
```
FedeDP, line 318: We do not allow to switch on llvm/clang version anymore, as it really makes no sense.

pkg/driverbuilder/builder/archlinux.go
```diff
@@ -23,7 +23,6 @@ type archlinux struct {
 type archlinuxTemplateData struct {
 	commonTemplateData
 	KernelDownloadURL string
-	GCCVersion        string
```
FedeDP, line 26: GCCVersion is not embed in `commonTemplateData` and automatically selected by a not-so-smart algorithm, when GCCVersionRequestor interface is not implemented by a builder.

pkg/driverbuilder/builder/archlinux.go
```diff
@@ -39,10 +38,14 @@ func (c archlinux) URLs(cfg Config, kr kernelrelease.KernelRelease) ([]string, e
 
 	if kr.Architecture == "amd64" {
 		urls = append(urls, fmt.Sprintf(
-			"https://archive.archlinux.org/packages/l/linux-headers/linux-headers-%s.%s-%s-%s.pkg.tar.xz",
+			"https://archive.archlinux.org/packages/l/linux-headers/linux-headers-%s.%s-%s.pkg.tar.xz",
+			kr.Fullversion,
+			kr.Extraversion,
+			kr.Architecture.ToNonDeb()))
+		urls = append(urls, fmt.Sprintf(
+			"https://archive.archlinux.org/packages/l/linux-headers/linux-headers-%s.%s-%s.pkg.tar.zst",
```
FedeDP, line 46: Support new zstd format for archlinux packages.

pkg/driverbuilder/builder/builders.go
```diff
@@ -63,6 +68,8 @@ type MinimumURLsBuilder interface {
 }
 
 func Script(b Builder, c Config, kr kernelrelease.KernelRelease) (string, error) {
+	c.SetGCCVersion(b, kr)
```
FedeDP, line 71: This sets the gcc version for the current build, given the builder (if it implements `GCCVersionRequestor` interface) or the kernelrelease.

pkg/driverbuilder/builder/builders.go
```diff
@@ -106,6 +113,132 @@ func Script(b Builder, c Config, kr kernelrelease.KernelRelease) (string, error)
...
 
+type GCCVersionRequestor interface {
+	GCCVersion(kr kernelrelease.KernelRelease) float64
+}
+
+type Image struct {
+	GCCVersion []float64
+}
+
+var images = map[string]Image{
```
FedeDP, line 124: Static map of images; when adding a new builder image, make sure to update this map.

pkg/driverbuilder/builder/flatcar.go
```diff
@@ -71,13 +71,20 @@ func (f flatcar) TemplateData(c Config, _ kernelrelease.KernelRelease, urls []st
...
 		commonTemplateData: c.toTemplateData(),
 		KernelDownloadURL:  urls[0],
-		GCCVersion:         flatcarGccVersion(f.info.GCCVersion),
 		FlatcarVersion:     f.version,
 		FlatcarChannel:     f.info.Channel,
 		KernelConfigURL:    f.kconfUrls[0],
 	}
 }
 
+func (f flatcar) GCCVersion(_ kernelrelease.KernelRelease) float64 {
```
FedeDP, line 80: Flatcar implements `GCCVersionRequestor` interface!

pkg/driverbuilder/builder/templates/amazonlinux.sh
```diff
@@ -28,7 +28,7 @@ mv usr/src/kernels/*/* /tmp/kernel
 # Build the kernel module
 cd {{ .DriverBuildDir }}
 
-make KERNELDIR=/tmp/kernel CC=/usr/bin/gcc LD=/usr/bin/ld.bfd CROSS_COMPILE=""
+make KERNELDIR=/tmp/kernel CC=/usr/bin/gcc-{{ .GCCVersion }} LD=/usr/bin/ld.bfd CROSS_COMPILE=""
```
FedeDP, line 31: We do always have a `.GCCVersion` in template now. Use it everywhere.

pkg/driverbuilder/builder/templates/amazonlinux.sh
```diff
@@ -37,6 +37,6 @@ modinfo {{ .ModuleFullPath }}
 {{ if .BuildProbe }}
 # Build the eBPF probe
 cd {{ .DriverBuildDir }}/bpf
-make LLC=/usr/bin/llc-{{ .LLVMVersion }} CLANG=/usr/bin/clang-{{ .LLVMVersion }} KERNELDIR=/tmp/kernel
+make KERNELDIR=/tmp/kernel
```
FedeDP, line 40: We always use default llc and clang now, avoid specifying them.

pkg/driverbuilder/builder/ubuntu.go
```diff
@@ -75,10 +74,29 @@ func (v ubuntu) TemplateData(c Config, kr kernelrelease.KernelRelease, urls []st
 		KernelDownloadURLS:   urls,
 		KernelLocalVersion:   kr.FullExtraversion,
 		KernelHeadersPattern: headersPattern,
-		GCCVersion:           ubuntuGCCVersionFromKernelRelease(kr),
 	}
 }
 
+func (v ubuntu) GCCVersion(kr kernelrelease.KernelRelease) float64 {
```
FedeDP, line 80: Ubuntu implements `GCCVersionRequestor` interface!

pkg/driverbuilder/builder/ubuntu_test.go
```diff
@@ -52,23 +52,23 @@ var tests = []struct {
...
-			gccVersion:  "8",
-			firstExtra:  "1129",
+			headersURLs: []string{"http://ports.ubuntu.com/ubuntu-ports/pool/main/l/linux-aws/linux-headers-4.15.0-1140-aws_4.15.0-1140.151_arm64.deb", "http://ports.ubuntu.com/ubuntu-ports/pool/main/l/linux-aws/linux-aws-headers-4.15.0-1140_4.15.0-1140.151_all.deb"},
```
FedeDP, line 68: This fix is also part of #201

pkg/kernelrelease/kernelrelease.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 var (
-	kernelVersionPattern = regexp.MustCompile(`(?P<fullversion>^(?P<version>0|[1-9]\d*)\.(?P<patchlevel>0|[1-9]\d*)\.(?P<sublevel>0|[1-9]\d*))(?P<fullextraversion>-(?P<extraversion>0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-_]*))*)?(\+[0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*)?$`)
+	kernelVersionPattern = regexp.MustCompile(`(?P<fullversion>^(?P<version>0|[1-9]\d*)\.(?P<patchlevel>0|[1-9]\d*)\.(?P<sublevel>0|[1-9]\d*))(?P<fullextraversion>[-|.](?P<extraversion>0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-_]*))*)?(\+[0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*)?$`)
```
FedeDP, line 12: Support kernelreleases like the archlinux one, ie:
`5.19.3-arch1-1` (note the `-arch1`)

pkg/kernelrelease/kernelrelease_test.go
```diff
@@ -144,6 +144,17 @@ func TestFromString(t *testing.T) {
 				FullExtraversion: "-1044-gke",
 			},
 		},
+		"arch version": {
```
FedeDP, line 147: Added a small test against archlinux kernelversion.

validate/isimagename.go
```diff
@@ -13,6 +13,11 @@ const alphabet = letters + digits + separators
 func isImageName(fl validator.FieldLevel) bool {
 	name := fl.Field().String()
 
+	// default value
+	if name == "" {
```
FedeDP, line 17: CustomBuilderImage is by default empty.
Allow it as a value, as we have a proper fallback in any case.

cmd/root_options.go
```diff
@@ -36,12 +36,6 @@ func init() {
 	validate.V.RegisterStructValidation(RootOptionsLevelValidation, RootOptions{})
 }
 
-func (ro *RootOptions) SetDefaults() {
```
FedeDP, line 39: Do not default the CustomBuilderImage value to builder.BaseImage.
Leave it empty when unset as we have a proper fallback.

pkg/driverbuilder/builder/builders.go
```diff
@@ -106,6 +113,132 @@ func Script(b Builder, c Config, kr kernelrelease.KernelRelease) (string, error)
...
+	} else {
+		targetGCC = defaultGCC(kr)
+	}
+
+	for name, img := range images {
+		var foundGCC float64
+		d := math.MaxFloat64
+		for _, gcc := range img.GCCVersion {
+			if math.Abs(gcc-targetGCC) < d {
+				// Find the nearest to targetGCC
```
dwindsor, line 180: Hmm, shouldn't this be the next available `gcc` version, rather than the nearest? For instance, if we have a `targetGCC` of 5.1, I'd think that 6 would be the version that should get chosen rather than 5?

pkg/driverbuilder/builder/ubuntu.go
dwindsor: Much cleaner to have this interface than e.g. `ubuntuGccVersionFromKernelRelease` =)

pkg/driverbuilder/builder/centos.go
```diff
@@ -23,7 +23,6 @@ type centos struct {
 type centosTemplateData struct {
 	commonTemplateData
 	KernelDownloadURL string
-	GCCVersion        string
```
dwindsor, line 26: Will we eventually be implementing `GCCVersionRequestor` interfaces for all builders? +1  from me for requiring `GCCVersionRequestor` to be defined for all new builders (moving forward, of course).

dwindsor (commented):
Thanks for the massive amount of planning/discussion/time that went into this! Not only does this MR allow us to support the latest drivers that upstream distros have to offer, it turns out that having an algorithm to select default GCC/builder versions greatly simplifies many other parts of `driverkit`. Well done! 

Long-term maintenance of this feature IIUC will include:
1. Updating each builder's `GccVersionRequestor` interface as new distro and gcc versions are released.
2. Updating `builders/builder.go`:
  a. `var images` mapping of docker images to gcc versions as new docker or gcc versions are released.
  b. `func defaultGcc` mapping of kernel major versions to gcc versions as new kernel or gcc versions are released.

pkg/kernelrelease/kernelrelease.go
dwindsor: `kernelVersionPattern` grows yet bigger =)

pkg/driverbuilder/builder/builders.go
```diff
@@ -106,6 +113,132 @@ func Script(b Builder, c Config, kr kernelrelease.KernelRelease) (string, error)
...
+	},
+	"bullseye": {
+		GCCVersion: []float64{9, 10},
+	},
+	"bookworm": {
+		GCCVersion: []float64{11, 12},
+	},
+}
+
+func defaultGCC(kr kernelrelease.KernelRelease) float64 {
```
dwindsor, line 136: Nit: might want a test for this function

EXONER4TED:
Hey, this PR is incredible/exactly what we need downstream too.

>Note that i completely dropped support for multiple clang versions, removing clang from our support matrix; if an eBPF probe cannot be built with clang shipped in any of the supported builder distros, then it is a bug of libs and an issue should be opened.

💯 

Seeing all of the standard packages used in the Docker images makes me happy 🙂

pkg/driverbuilder/builder/README.md
```diff
@@ -124,23 +123,48 @@ Indeed, the hardest part is fetching the kernel headers urls for each distro.
...
-* llvm-7
-* llvm-12
+* /usr/bin/clang (clang-14)
 
-You can dynamically choose the one you prefer, likely switching on the kernel version.  
-For an example, you can check out Debian builder, namely: `debianLLVMVersionFromKernelRelease`.
+Note, however, that there is no mechanism to dynamically choose a clang version,  
+as changing it should not be needed.
+The build will use the one provided by the chosen builder image.  
+Any failure must be treated as a bug, therefore and issue must be opened on [libs](https://github.com/falcosecurity/libs) repository.
```
EXONER4TED, line 167: 😍

pkg/driverbuilder/builder/builders.go
EXONER4TED: I second this question

EXONER4TED:
Aside from @dwindsor's question on nearest GCC version, I say :shipit: - this is a *gamechanger* for us downstream where we're looking at having multiple build images around to handle new stuff coming in. 

Incredible work 🎉

pkg/driverbuilder/builder/builders.go
FedeDP: Yeah, i am not sure if it makes any difference: i think that a target gcc of let's say 5.4 will always be buildable from a gcc 5.0 let's say.
But we can easily make the change here if you wish :)

pkg/driverbuilder/builder/centos.go
FedeDP: When not defined, `builder.defaultGCCVersion` will give us the GCC version to be used using the kernelrelease.
I think indeed we should need to implement `GCCVersionRequestor` only in very specific cases, ie: when the default algorithm is off for some reasons.

pkg/driverbuilder/builder/builders.go
FedeDP: Agree!

FedeDP:
Thank you both! @EXONER4TED @dwindsor :)

I think i will:
* move the builder.README in a doc file under docs
* add a new doc file about how to add a new builder image
* add a test for `defaultGCC` function
* Port the `defaultGCC` algorithm to use the nearest greater gcc version instead of the nearest one (?, i am not sure about this one)

pkg/driverbuilder/builder/builders.go
FedeDP: I wanted to avoid cases where a GCC X is too new to build a GCC Y target, ie:
* target is 6.5
* we have 6.3 or 9
* we chose 9
* 9 does not build because it is too new (glibc mismatch)

FedeDP:
I think that to fix tests, we could update this PR using `fededp` as docker org for the `build-test` CI jobs; this would allow the build CI job to pass and the new latest img to be pushed to `falcosecurity` docker org once the PR is merged.
Or, we could also entirely disable the integration tests (that require proper docker images) just for this and then open a new PR to add them back, once the images are pushed to falcosecurity.

dwindsor:
> Or, we could also entirely disable the integration tests (that require proper docker images) just for this and then open a new PR to add them back, once the images are pushed to falcosecurity.

tbh I think this is the cleanest way forward. wdyt?

pkg/driverbuilder/builder/builders.go
EXONER4TED: That makes sense to me, thank you for explaining!

FedeDP:
Btw to test this out:
* update `GetBuilderImage` to just return image for your arch (because with next step we will just push the image for current arch, not the manifest one with both arm64 and x86_64): assuming you are on `x86_64`, add "_x86_64" as a suffix to https://github.com/falcosecurity/driverkit/blob/new/multiple_builders/pkg/driverbuilder/builder/builders.go#L235
* call `push/builder` makefile target pointing the env variable `DOCKER_ORG`(https://github.com/falcosecurity/driverkit/blob/master/Makefile#L19) to your personal dockerhub name, ie: `make -e DOCKER_ORG=fededp push/builder`

* ok now you are good to go!

FedeDP:
I will rebase on top of new master.

FedeDP:
Rebased :green_circle:

FedeDP:
Ok now ci is failing as i thought:
```
DEBU pulling builder image                         arch=amd64 image="docker.io/falcosecurity/driverkit-builder_buster:latest"
FATA exiting                                       error="Error response from daemon: pull access denied for falcosecurity/driverkit-builder_buster, repository does not exist or may require 'docker login': denied: requested access to the resource is denied"
make: *** [Makefile:127: test/x86_64/configs/al.yaml] Error 1
```

FedeDP:
As per https://github.com/falcosecurity/test-infra/pull/824 issues with various centos 3.X kernels, that need gcc 4.9, i added gcc-4.9 to the buster builder image.

leogr:
Great job :star_struck: 

:clap:

FedeDP:
Rebased on top of master ;)

dwindsor (approved):

PR #209, opened 2022-09-16
fix(pkg/driverbuilder): dynamically load driver makefile objects by downloading Makefile.in from libs github

FedeDP:
**What type of PR is this?**

/kind bug

/area pkg

To be able to be kept in sync with libs driver makefile, we download [the raw makefile](https://github.com/falcosecurity/libs/blob/master/driver/Makefile.in) for the required driverversion, and we load from it the required objects.

This is a bit ugly, because we will be going to download the entire repo at requested driverversion anyway, in each builder template.
Ideally, we'd want to move the repo download outside, load makefile objects, and share the volume with the builder image.

I will surely try to make this in a future PR.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/driverbuilder/templates.go
+++ b/pkg/driverbuilder/templates.go
@@ -1,7 +1,11 @@
 package driverbuilder
 
 import (
+	"fmt"
+	"github.com/falcosecurity/driverkit/pkg/driverbuilder/builder"
 	"io"
+	"net/http"
+	"strings"
 	"text/template"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/driverbuilder/templates.go
```diff
@@ -64,6 +69,27 @@ func renderMakefile(w io.Writer, md makefileData) error {
 	return t.Execute(w, md)
 }
 
+func LoadMakefileObjList(c builder.Config) (string, error) {
```
FedeDP, line 72: Get the raw content of makefile.in in libs repo at requested driverversion;
then, find the `@DRIVER_NAME@-y +=` row and take the second part (ie: the list of objects required for the module).

PR #215, opened 2022-09-23
fix(pkg,cmd,validate): fixed flatcar builder

FedeDP:
**What type of PR is this?**

/kind bug

/area pkg

Fix a crash in the flatcar builder; moreover, make it actually work.
Basically it is now the same as vanilla/minikube builders, with some small differences. It will now require `kernelconfigdata` option. That is much more smart than fetching the kernelconfigdata URL only to download it in the template script, considering that we already had a `kernelconfigdata` option in driverkit.

Moreover, moved all builder structs receivers to pointers, so that we can actually update any struct members in receiver methods.

Finally, actually require `kernelconfigdata` param for minikube and flatcar too.

This goes together with https://github.com/falcosecurity/kernel-crawler/pull/50.

Flatcar crash + make it working.

Fixes #

```release-note
fix(pkg): fixed flatcar builder crash and implementation
```

### DIFF
--- a/cmd/root_options.go
+++ b/cmd/root_options.go
@@ -119,8 +119,12 @@ func (ro *RootOptions) toBuild() *builder.Build {
 func RootOptionsLevelValidation(level validator.StructLevel) {
 	opts := level.Current().Interface().(RootOptions)
 
-	if len(opts.KernelConfigData) == 0 && opts.Target == builder.TargetTypeVanilla.String() {
-		level.ReportError(opts.KernelConfigData, "kernelConfigData", "KernelConfigData", "required_kernelconfigdata_with_target_vanilla", "")
+	if opts.Target == builder.TargetTypeVanilla.String() ||
+		opts.Target == builder.TargetTypeMinikube.String() ||
+		opts.Target == builder.TargetTypeFlatcar.String() {
+		if len(opts.KernelConfigData) == 0 {
+			level.ReportError(opts.KernelConfigData, "kernelConfigData", "KernelConfigData", "required_kernelconfigdata_with_target_vanilla", "")
+		}
 	}
 
 	if opts.KernelVersion == "" && (opts.Target == builder.TargetTypeUbuntu.String()) {
--- a/pkg/driverbuilder/builder/amazonlinux.go
+++ b/pkg/driverbuilder/builder/amazonlinux.go
@@ -62,26 +62,26 @@ type amazonlinuxTemplateData struct {
 	KernelDownloadURLs []string
 }
 
-func (a amazonlinux) Name() string {
+func (a *amazonlinux) Name() string {
 	return TargetTypeAmazonLinux.String()
 }
 
-func (a amazonlinux) TemplateScript() string {
+func (a *amazonlinux) TemplateScript() string {
 	return amazonlinuxTemplate
 }
 
-func (a amazonlinux) URLs(_ Config, kr kernelrelease.KernelRelease) ([]string, error) {
+func (a *amazonlinux) URLs(_ Config, kr kernelrelease.KernelRelease) ([]string, error) {
 	return fetchAmazonLinuxPackagesURLs(a, kr)
 }
 
-func (a amazonlinux) TemplateData(c Config, kr kernelrelease.KernelRelease, urls []string) interface{} {
+func (a *amazonlinux) TemplateData(c Config, kr kernelrelease.KernelRelease, urls []string) interface{} {
 	return amazonlinuxTemplateData{
-		commonTemplateData: c.toTemplateData(),
+		commonTemplateData: c.toTemplateData(a, kr),
 		KernelDownloadURLs: urls,
 	}
 }
 
-func (a amazonlinux) repos() []string {
+func (a *amazonlinux) repos() []string {
 	return []string{
 		"latest/updates",
 		"latest/main",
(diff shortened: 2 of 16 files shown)


cmd/root_options.go
```diff
@@ -119,8 +119,12 @@ func (ro *RootOptions) toBuild() *builder.Build {
 func RootOptionsLevelValidation(level validator.StructLevel) {
 	opts := level.Current().Interface().(RootOptions)
 
-	if len(opts.KernelConfigData) == 0 && opts.Target == builder.TargetTypeVanilla.String() {
-		level.ReportError(opts.KernelConfigData, "kernelConfigData", "KernelConfigData", "required_kernelconfigdata_with_target_vanilla", "")
+	if opts.Target == builder.TargetTypeVanilla.String() ||
```
FedeDP, line 122: Validate: we need kernelconfigdata for { vanilla, minikube, flatcar } builders.

pkg/driverbuilder/builder/amazonlinux.go
```diff
@@ -62,26 +62,26 @@ type amazonlinuxTemplateData struct {
...
-func (a amazonlinux) URLs(_ Config, kr kernelrelease.KernelRelease) ([]string, error) {
+func (a *amazonlinux) URLs(_ Config, kr kernelrelease.KernelRelease) ([]string, error) {
 	return fetchAmazonLinuxPackagesURLs(a, kr)
 }
 
-func (a amazonlinux) TemplateData(c Config, kr kernelrelease.KernelRelease, urls []string) interface{} {
+func (a *amazonlinux) TemplateData(c Config, kr kernelrelease.KernelRelease, urls []string) interface{} {
 	return amazonlinuxTemplateData{
-		commonTemplateData: c.toTemplateData(),
+		commonTemplateData: c.toTemplateData(a, kr),
```
FedeDP, line 79: `c.toTemplateData(a, kr)` now internally calls `c.setGCCVersion`, that's why it needs both the kernelrelease and the builder now.

pkg/driverbuilder/builder/builders.go
```diff
@@ -105,6 +103,10 @@ func Script(b Builder, c Config, kr kernelrelease.KernelRelease) (string, error)
 	}
 
 	td := b.TemplateData(c, kr, urls)
+	if tdErr, ok := td.(error); ok {
```
FedeDP, line 106: TemplateData can now return an interface that resolves to an error, making `Script` exit with an error.

pkg/driverbuilder/builder/flatcar.go
```diff
@@ -23,66 +22,68 @@ func init() {
...
-		return nil, err
-	}
-
-	f.kconfUrls, err = getResolvingURLs(fetchFlatcarKernelConfigURL(kr.Architecture, f.info.Channel, kr.Fullversion))
-	if err != nil {
+func (f *flatcar) URLs(_ Config, kr kernelrelease.KernelRelease) ([]string, error) {
+	if err := f.fillFlatcarInfos(kr); err != nil {
 		return nil, err
 	}
 	return fetchFlatcarKernelURLS(f.info.KernelVersion), nil
```
FedeDP, line 44: Note: we pass the resolved kernel version because flatcar takes as `kernelrelease` config value the flatcar version (like 1192), not the kernel one.

pkg/driverbuilder/builder/flatcar.go
```diff
@@ -23,66 +22,68 @@ func init() {
...
 }
 
-func (f flatcar) TemplateData(c Config, _ kernelrelease.KernelRelease, urls []string) interface{} {
+func (f *flatcar) TemplateData(c Config, kr kernelrelease.KernelRelease, urls []string) interface{} {
+	// This happens when `kernelurls` option is passed,
+	// therefore URLs() method is not called.
+	if f.info == nil {
```
FedeDP, line 50: Avoid dereferencing a nil pointer. We need the target GCC to be loaded in any case.

Moreover, by moving the `setGCCVersion` call inside `toTemplateData` we enforce the GCC version to be set last, after the builder has done everything it needs to eventually customize GCCVersion (in this case, f.info loading).

pkg/driverbuilder/builder/flatcar.go
```diff
@@ -23,66 +22,68 @@ func init() {
...
 	}
 }
 
-func (f flatcar) GCCVersion(_ kernelrelease.KernelRelease) float64 {
-	v, err := strconv.ParseFloat(f.info.GCCVersion, 64)
-	if err != nil {
-		return 8
+func (f *flatcar) GCCVersion(_ kernelrelease.KernelRelease) float64 {
+	// Parse maj,min,patch
+	gccV := kernelrelease.FromString(f.info.GCCVersion)
```
FedeDP, line 64: GCCVersion can be something like `8.3.0`, we cannot parse it as float.
Instead, parse it as a semver kernelrelease version.

TODO: we should really:
* rename kernelrelease.FromString to something like `semver.FromString`
* make all the target GCC logic to use semver.SemVer struct instead of float64

I will make this in a subsequent PR.

pkg/driverbuilder/builder/flatcar.go
```diff
@@ -143,15 +144,6 @@ func fetchFlatcarPackageListURL(architecture kernelrelease.Architecture, flatcar
 	return urls
 }
 
-func fetchFlatcarKernelConfigURL(architecture kernelrelease.Architecture, flatcarChannel, flatcarVersion string) []string {
```
FedeDP, line 146: We don't need to fetch kernel confgi URL anymore, because we expect kernelconfig to be already passed as `kernelconfigdata` option.

pkg/driverbuilder/builder/templates/flatcar.sh
```diff
@@ -20,9 +20,10 @@ rm -Rf /tmp/kernel
 mkdir -p /tmp/kernel
 mv /tmp/kernel-download/*/* /tmp/kernel
 
-curl --silent -o /tmp/kernel.config -SL {{ .KernelConfigURL }}
```
FedeDP, line 23: No need to download the kernel config file. We already have it passed by user/config ;)

dwindsor (approved):

PR #217, opened 2022-09-29
chore: use semver package for gcc versions and kernel versions.

FedeDP:
**What type of PR is this?**

/kind cleanup

/area cmd

/area pkg

/area tests

This PR enforces the use of a semver library to manage:
* gcc versions
* kernel versions

Moreover, it updates and fixes an ubuntu test (as always, the ubuntu header for the old test-case was no more there; just updated the test version)

Fixes #

```release-note
NONE
```

### DIFF
--- a/cmd/root_options.go
+++ b/cmd/root_options.go
@@ -31,7 +31,7 @@ type RootOptions struct {
 	Target           string   `validate:"required,target" name:"target"`
 	KernelConfigData string   `validate:"omitempty,base64" name:"kernel config data"` // fixme > tag "name" does not seem to work when used at struct level, but works when used at inner level
 	BuilderImage     string   `validate:"omitempty,imagename" name:"builder image"`
-	GCCVersion       string   `validate:"omitempty,semver" name:"gcc version"`
+	GCCVersion       string   `validate:"omitempty,semvertolerant" name:"gcc version"`
 	KernelUrls       []string `name:"kernel header urls"`
 	Repo             RepoOptions
 	Output           OutputOptions
--- a/cmd/root_options.go
+++ b/cmd/root_options.go
@@ -30,8 +30,8 @@ type RootOptions struct {
 	KernelRelease    string   `validate:"required,ascii" name:"kernel release"`
 	Target           string   `validate:"required,target" name:"target"`
 	KernelConfigData string   `validate:"omitempty,base64" name:"kernel config data"` // fixme > tag "name" does not seem to work when used at struct level, but works when used at inner level
-	BuilderImage     string   `validate:"imagename" name:"builder image"`
-	GCCVersion       float64  `name:"gcc version"`
+	BuilderImage     string   `validate:"omitempty,imagename" name:"builder image"`
+	GCCVersion       string   `validate:"omitempty,semver" name:"gcc version"`
 	KernelUrls       []string `name:"kernel header urls"`
 	Repo             RepoOptions
 	Output           OutputOptions
--- a/pkg/driverbuilder/builder/builders.go
+++ b/pkg/driverbuilder/builder/builders.go
@@ -4,9 +4,9 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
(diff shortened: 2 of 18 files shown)


FedeDP:
NOTE: this is far less elegant than what i hoped :/ 
I am sorry for the `builders.go` part, but we needed a way to map a semver to a proper string (that is, the gcc string actually installed in the builder image); basically, we couldn't use `semver.Version` for builder images spec, because then we would have lost the mapping between 
```
v := semver.Version { Major: 5, Minor: 3 }
v.String() -> "5.3.0"
```
and real gcc name in the system (ie: `gcc-5.3`).
While we could possibly just say "skip 0es" writing an ad hoc function, i think that was not a very future proof way (what if a certain point we would have needed like "5.3.0" for real?).

In the end, i chose to keep builder images mapping using string for the provided gcc versions, and parse them automatically to `semver.Version` in the gcc selection algorithm.

cmd/root_options.go
```diff
@@ -30,8 +30,8 @@ type RootOptions struct {
 	KernelRelease    string   `validate:"required,ascii" name:"kernel release"`
 	Target           string   `validate:"required,target" name:"target"`
 	KernelConfigData string   `validate:"omitempty,base64" name:"kernel config data"` // fixme > tag "name" does not seem to work when used at struct level, but works when used at inner level
-	BuilderImage     string   `validate:"imagename" name:"builder image"`
-	GCCVersion       float64  `name:"gcc version"`
+	BuilderImage     string   `validate:"omitempty,imagename" name:"builder image"`
```
FedeDP, line 33: Do not validate empty BuilderImage. This allowed us to skip a small check in the `imagename` validator.

pkg/driverbuilder/builder/builders.go
```diff
@@ -116,97 +116,145 @@ func Script(b Builder, c Config, kr kernelrelease.KernelRelease) (string, error)
...
 	case 2:
-		return 4.8
+		return semver.Version{Major: 4, Minor: 8}
 	default:
-		return 12
+		return semver.Version{Major: 12}
 	}
 }
 
+func mustParseTolerant(gccStr string) semver.Version {
```
FedeDP, line 173: `ParseTolerant` allows to parse eg: `5` to a `5.0.0` semver. Using just `Parse`, it would panic.

pkg/driverbuilder/builder/builders.go
```diff
@@ -116,97 +116,145 @@ func Script(b Builder, c Config, kr kernelrelease.KernelRelease) (string, error)
...
+		panic(err)
+	}
+	return g
+}
+
+// Simple algorithm:
+// sort versions passed as first params,
+// then, find the higher gcc version which is also lower than target.
+// Moreover, try harder to keep same major gcc version.
+func findNearestGCCVersion(gccs []semver.Version, target semver.Version) semver.Version {
```
FedeDP, line 185: This is the core method for gcc version selector. Please give it a proper look :)

pkg/kernelrelease/kernelrelease.go
```diff
@@ -65,13 +66,11 @@ func (a Architecture) String() string {
...
 // (it it set for builders in kernelReleaseFromBuildConfig())
 type KernelRelease struct {
-	Fullversion      string       `json:"full_version"`
-	Version          int          `json:"version"`
-	PatchLevel       int          `json:"patch_level"`
-	Sublevel         int          `json:"sublevel"`
-	Extraversion     string       `json:"extra_version"`
-	FullExtraversion string       `json:"full_extra_version"`
-	Architecture     Architecture `json:"architecture"`
+	Fullversion string
```
FedeDP, line 69: Dropped json tags: do we ever needed them?

dwindsor:
> I am sorry for the builders.go part, but we needed a way to map a semver to a proper string (that is, the gcc string actually installed in the builder image); 

Not a huge deal, the switch to semver is worth it imo

PR #224, opened 2022-10-19
Adjust `headersPattern` Detection for Ubuntu

### LINKED ISSUE #223: Ubuntu lowlatency-hwe's not working
**Describe the bug**
Ubuntu `lowlatency-hwe-*` kernels don't work with the current Ubuntu template script. Example:
```
DEBU + curl --silent -o kernel.deb -SL http://security.ubuntu.com/ubuntu/pool/main/l/linux-lowlatency-hwe-5.15/linux-headers-5.15.0-33-lowlatency_5.15.0-33.34~20.04.1_amd64.deb
DEBU + ar x kernel.deb
DEBU + tar -xf data.tar.xz
DEBU �+ curl --silent -o kernel.deb -SL http://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux-lowlatency-hwe-5.15/linux-lowlatency-hwe-5.15-headers-5.15.0-33_5.15.0-33.34~20.04.1_all.deb
DEBU + ar x kernel.deb
DEBU + tar -xf data.tar.xz
DEBU �+ curl --silent -o kernel.deb -SL http://security.ubuntu.com/ubuntu/pool/main/l/linux-lowlatency-hwe-5.15/linux-lowlatency-hwe-5.15-headers-5.15.0-33_5.15.0-33.34~20.04.1_all.deb
DEBU + ar x kernel.deb
DEBU + tar -xf data.tar.xz
DEBU �+ curl --silent -o kernel.deb -SL http://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux-lowlatency-hwe-5.15/linux-headers-5.15.0-33-lowlatency_5.15.0-33.34~20.04.1_amd64.deb
DEBU + ar x kernel.deb
DEBU + tar -xf data.tar.xz
DEBU #+ cd /tmp/kernel-download/usr/src/
DEBU 7++ find . -type d -name 'linux-headers*lowlatency-hwe'
DEBU "++ head -n 1
DEBU ++ xargs readlink -f
DEBU Freadlink: missing operand
```
The `find . -type d -name 'linux-headers*lowlatency-hwe'` does not work as the `headersPattern` needs to actually be `linux-headers*lowlatency`, not `linux-headers*lowlatency-hwe`. I'm not sure why, the directory extracted out of the deb just doesn't have the `-hwe` on the end.

It should be possible to fix this by just splitting the `flavor` on `-` and taking the `split[0]` element for the `headersPattern`.

**How to reproduce it**

Example command (kernel urls come from kernel crawler):
```
_output/bin/driverkit docker --output-module ./test-module.ko --kernelrelease 5.15.0-33-lowlatency-hwe-5.15 --kernelversion 34~20.04.1 --driverversion 2.0.0+driver --target ubuntu --moduledevicename scwx-falco --moduledrivername scwx-falco --kernelurls http://security.ubuntu.com/ubuntu/pool/main/l/linux-lowlatency-hwe-5.15/linux-headers-5.15.0-33-lowlatency_5.15.0-33.34\~20.04.1_amd64.deb --kernelurls http://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux-lowlatency-hwe-5.15/linux-lowlatency-hwe-5.15-headers-5.15.0-33_5.15.0-33.34\~20.04.1_all.deb --kernelurls http://security.ubuntu.com/ubuntu/pool/main/l/linux-lowlatency-hwe-5.15/linux-lowlatency-hwe-5.15-headers-5.15.0-33_5.15.0-33.34\~20.04.1_all.deb --kernelurls http://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux-lowlatency-hwe-5.15/linux-headers-5.15.0-33-lowlatency_5.15.0-33.34\~20.04.1_amd64.deb -l debug
```

**Expected behaviour**
That it builds :)

**Screenshots**

**Environment**

- Falco version:

- System info:

- Cloud provider or hardware configuration:
- OS:

- Kernel:

- Installation method:

**Additional context**

EXONER4TED:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area cmd

/area pkg

> /area docs

> /area tests

Adjusting the `headersPattern` for Ubuntu. See #223 

Splitting the `flavor` on `-` should return just the first part of the flavor, which will still work for the `headersPattern`. If the `flavor` does not contain `-`, then it should just return the original `flavor` like normal.

Fixes #223 

```release-note

```

### DIFF
--- a/pkg/driverbuilder/builder/templates/ubuntu.sh
+++ b/pkg/driverbuilder/builder/templates/ubuntu.sh
@@ -22,6 +22,7 @@ tar -xf data.tar.*
 {{end}}
 
 cd /tmp/kernel-download/usr/src/
+ls -altr
 sourcedir=$(find . -type d -name "{{ .KernelHeadersPattern }}" | head -n 1 | xargs readlink -f)
 
 {{ if .BuildModule }}
--- a/pkg/driverbuilder/builder/ubuntu.go
+++ b/pkg/driverbuilder/builder/ubuntu.go
@@ -59,7 +59,7 @@ func (v *ubuntu) TemplateData(c Config, kr kernelrelease.KernelRelease, urls []s
 	if flavor == "hwe" {
 		headersPattern = "linux-headers*generic"
 	} else {
-		// some flavors (ex: lowlatency-hwe) only contain the first part of the flavor in the  directory extracted from the .deb
+		// some flavors (ex: lowlatency-hwe) only contain the first part of the flavor in the directory extracted from the .deb
 		// splitting a flavor without a "-" should just return the original flavor back	
 		headersPattern = fmt.Sprintf("linux-headers*%s", strings.Split(flavor, "-")[0])
 	}
--- a/pkg/driverbuilder/builder/ubuntu.go
+++ b/pkg/driverbuilder/builder/ubuntu.go
@@ -59,8 +59,7 @@ func (v *ubuntu) TemplateData(c Config, kr kernelrelease.KernelRelease, urls []s
 	if flavor == "hwe" {
 		headersPattern = "linux-headers*generic"
 	} else {
-		// some flavors (ex: lowlatency-hwe) only contain the first part of the flavor in the
-		// directory extracted from the .deb
+		// some flavors (ex: lowlatency-hwe) only contain the first part of the flavor in the  directory extracted from the .deb
 		// splitting a flavor without a "-" should just return the original flavor back	
 		headersPattern = fmt.Sprintf("linux-headers*%s", strings.Split(flavor, "-")[0])
 	}
--- a/pkg/driverbuilder/builder/ubuntu.go
+++ b/pkg/driverbuilder/builder/ubuntu.go
@@ -59,7 +59,10 @@ func (v *ubuntu) TemplateData(c Config, kr kernelrelease.KernelRelease, urls []s
 	if flavor == "hwe" {
 		headersPattern = "linux-headers*generic"
 	} else {
-		headersPattern = fmt.Sprintf("linux-headers*%s", flavor)
+		// some flavors (ex: lowlatency-hwe) only contain the first part of the flavor in the
+		// directory extracted from the .deb
+		// splitting a flavor without a "-" should just return the original flavor back	
+		headersPattern = fmt.Sprintf("linux-headers*%s", strings.Split(flavor, "-")[0])
 	}
 
 	return ubuntuTemplateData{
--- a/pkg/driverbuilder/builder/ubuntu_test.go
+++ b/pkg/driverbuilder/builder/ubuntu_test.go
@@ -229,7 +229,7 @@ var tests = []struct {
 			flavor      string
 			err         error
 		}{
-			headersURLs: []string{"https://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux-kvm/linux-headers-5.19.0-1006-kvm_5.19.0-1006.6_amd64.deb", "https://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux-kvm/linux-kvm-headers-5.19.0-1006_5.19.0-1006.6_all.deb"},
+			headersURLs: []string{},
 			urls:        []string{"https://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux/linux-headers-5.19.0-1006-kvm_5.19.0-1006.6_amd64_all.deb", "https://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux/linux-headers-5.19.0-1006-kvm_5.19.0-1006.6_amd64.deb", "https://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux/linux-kvm-headers-5.19.0-1006_5.19.0-1006.6_all.deb", "https://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux-kvm/linux-headers-5.19.0-1006-kvm_5.19.0-1006.6_amd64_all.deb", "https://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux-kvm/linux-headers-5.19.0-1006-kvm_5.19.0-1006.6_amd64.deb", "https://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux-kvm/linux-kvm-headers-5.19.0-1006_5.19.0-1006.6_all.deb", "https://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux-kvm-5.19/linux-headers-5.19.0-1006-kvm_5.19.0-1006.6_amd64_all.deb", "https://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux-kvm-5.19/linux-headers-5.19.0-1006-kvm_5.19.0-1006.6_amd64.deb", "https://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux-kvm-5.19/linux-kvm-headers-5.19.0-1006_5.19.0-1006.6_all.deb"},
 			gccVersion: semver.Version{
 				Major: 12,


FedeDP:
Hi Logan! This LGTM, i reserve a better look tomorrow though :) 
On a side note, i wanted to ask you if you considered joining David and me as maintainers of driverkit! I think you are doing a great job and would love to get you on board!

pkg/driverbuilder/builder/templates/ubuntu.sh
```diff
@@ -22,6 +22,7 @@ tar -xf data.tar.*
 {{end}}
 
 cd /tmp/kernel-download/usr/src/
+ls -altr
```
dwindsor, line 25: Is this needed?

EXONER4TED:
Oh cool! David had mentioned it to me, but personal life got a little crazy for a while. I'd be happy to join now though that things have calmed down 🙂 how do I go about doing that?

pkg/driverbuilder/builder/templates/ubuntu.sh
EXONER4TED: I added it for useful debugging since it will only show in debug mode. I can remove it if you'd like. It would just have to be added back for debugging again later haha

dwindsor: Oh I thought it might've just been mistakenly left over. Makes sense to have it in there.

EXONER4TED: Example:
```
DEBU .+ cd /tmp/kernel-download/usr/src/
DEBU + ls -altr
DEBU �total 16
DEBU drwxr-xr-x 24 root root 4096 Oct  5  2020 linux-headers-5.4.0-51
DEBU drwxr-xr-x  4 root root 4096 Oct  5  2020 ..
DEBU drwxr-xr-x  4 root root 4096 Oct  5  2020 .
DEBU drwxr-xr-x  7 root root 4096 Oct 19 21:07 linux-headers-5.4.0-51-generic
DEBU =++ find . -type d -name 'linux-headers*generic'
```
Tells you what's actually in the folder before the `find` runs... So if `find` doesn't work, you can at least see what was actually there in debug mode.

pkg/driverbuilder/builder/ubuntu.go
```diff
@@ -59,7 +59,9 @@ func (v *ubuntu) TemplateData(c Config, kr kernelrelease.KernelRelease, urls []s
 	if flavor == "hwe" {
 		headersPattern = "linux-headers*generic"
 	} else {
-		headersPattern = fmt.Sprintf("linux-headers*%s", flavor)
+		// some flavors (ex: lowlatency-hwe) only contain the first part of the flavor in the directory extracted from the .deb
+		// splitting a flavor without a "-" should just return the original flavor back	
+		headersPattern = fmt.Sprintf("linux-headers*%s", strings.Split(flavor, "-")[0])
```
dwindsor, line 64: Thanks for this MR! And welcome to Falco =)

IIUC, this assumes that the flavor will contain at most one dash ("-")?

FedeDP:
> how do I go about doing that?

A couple of things:
* you need to present yourself and propose yourself as a maintainer during the weekly community call 
* moreover, it would be great if you could come and partecipate to the weekly community call in general :) it helps the community growth, and helps you keeping in touch with other project's areas!

After the first point, you can then open a PR here on driverkit to add yourself to the OWNERS file; we will then leave it open for a week or so, to allow anyone who has any comment, to be able to make it.
If nobody raises any objections, we can then approve it and...welcome!

pkg/driverbuilder/builder/templates/ubuntu.sh
FedeDP, line 25: I think i was the one removing all of these :/ but yep you are right, they help with debugging (even if they generate so much noice). It's ok to add it back!

pkg/driverbuilder/builder/ubuntu.go
FedeDP: I guess it assumes that when there are dashes, only the first part is actually interesting.
A side note: wouldn't it be better if we were able to return a list of headersPattern and in the template script cycle through all of them and stop at the first matching one?
Just an idea...

pkg/driverbuilder/builder/ubuntu_test.go
```diff
@@ -229,7 +229,7 @@ var tests = []struct {
 			flavor      string
 			err         error
 		}{
-			headersURLs: []string{"https://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux-kvm/linux-headers-5.19.0-1006-kvm_5.19.0-1006.6_amd64.deb", "https://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux-kvm/linux-kvm-headers-5.19.0-1006_5.19.0-1006.6_all.deb"},
+			headersURLs: []string{},
```
FedeDP, line 232: Mmmh nice shortcut ahah but i think you need to update/fix headerURLs instead of truncating them :P 

Moreover, the ubuntu tests are really flaky because they depend upon packages provided by ubuntu (ie: headers), but ubuntu is pretty quick in upgrading its supported kernels, and like monthly these tests must be fixed...it would be great to find a way to fix them in a way that aren't going to break again.
I think they are useful but i don't like flaky tests.

EXONER4TED: You're right about it not being the right answer to test this way. The reason I truncated is because that kernel no longer exists (those URLs now produce 404s). It would be better to just remove the URL checks that are reliant on Ubuntu, but I figured that was outside the scope of this PR. I can go adjust all the tests too if you'd like.

pkg/driverbuilder/builder/templates/ubuntu.sh
EXONER4TED: As long as it's not `-v` on `tar`, I'm good for more debug commands haha. The `tar -xvf` is the massive one...

pkg/driverbuilder/builder/ubuntu.go
EXONER4TED: You're right - the idea was that only the first part of the flavor is interesting. Take this example here (which is driverkit building from the `test/` folder - an `ubuntu-generic`):
```
DEBU .+ cd /tmp/kernel-download/usr/src/
DEBU + ls -altr
DEBU �total 16
DEBU drwxr-xr-x 24 root root 4096 Oct  5  2020 linux-headers-5.4.0-51
DEBU drwxr-xr-x  4 root root 4096 Oct  5  2020 ..
DEBU drwxr-xr-x  4 root root 4096 Oct  5  2020 .
DEBU drwxr-xr-x  7 root root 4096 Oct 19 21:07 linux-headers-5.4.0-51-generic
DEBU =++ find . -type d -name 'linux-headers*generic'
```
There are 2 folders in `/tmp/kernel-download/usr/src`:
1. `linux-headers-5.4.0-51`
2. `linux-headers-5.4.0-51-generic`

My first thought was just to do `linux-headers*`, but that finds the incorrect folder. We need the script to find the `linux-headers*generic`.

In some cases though, the flavors for Ubuntu get weird, specifically `hwe` kernels. `hwe` kernels will be `hwe` in the URL path, but when extracted from the `.deb` show up as `-generic` lol. That's the reason for the first case:
```golang
	headersPattern := ""
	if flavor == "hwe" {
		headersPattern = "linux-headers*generic"
	} 
```

However, in other cases, such as `lowlatency-hwe` kernels, the directory extracted from the `.deb` just chops off the `-hwe` on the end. That produces errors in the current script:
```
DEBU #+ cd /tmp/kernel-download/usr/src/
DEBU 7++ find . -type d -name 'linux-headers*lowlatency-hwe'
DEBU "++ head -n 1
DEBU ++ xargs readlink -f
DEBU Freadlink: missing operand
```
`'linux-headers*lowlatency-hwe'` won't exist...

So my idea was to just chop off the first part of the flavor, regardless of `how-long-the-flavor-might-be` and search for `linux-headers*<first-part-of-flavor>`. That gets us enough of the flavor string to not match the `linux-headers*` directory, but still match the flavor directory.

Another way we could do this is just check for `lowlatency-hwe` kernels, or maybe just check for `-hwe` and handle it specifically.. I was hoping to avoid one off checks though and go for something generic.

We could also try to move this checking to the script 🤔 I'm unsure I like the idea of trying to list all the flavors possible and iterate over them... What we really need is to just check if the directory extracted has _a_ flavor at all in its name.

Let me know what you think we should do though. I'm not convinced what I did is perfect either haha.

EXONER4TED:
When is the Falco weekly call/how do I get an invite to that? I'd be happy to join!

pkg/driverbuilder/builder/ubuntu_test.go
FedeDP: I think we can do it in another PR, you are right! Thanks!

pkg/driverbuilder/builder/ubuntu.go
EXONER4TED: To answer @dwindsor's original question: it assumes that the only interesting part of the `flavor` when it comes to that `find` command in the script is the first word of the `flavor`.

So even if the flavor is `apple-orange-banana`, we only care about `apple` because the 2 directories extracted from the `.deb` should look like:
1. `linux-headers-<kernel-release>`
2. `linux-headers-<kernel-release>-apple-orange-banana`

FedeDP: I think your solution is good @EXONER4TED! Only time will tell :)

FedeDP:
@EXONER4TED you can find all the relevant info here: https://github.com/falcosecurity/community :)

pkg/driverbuilder/builder/ubuntu.go
dwindsor: Agreed, we're at the mercy of the vendor for how they'll change their naming schemes.

dwindsor (approved):

PR #226, opened 2022-10-20
update: avoid building unfeasible kmods and probes

jasondellaluce:
**What type of PR is this?**

/kind feature

/area cmd

/area pkg

This updates the tool to figure out if the kmod and the eBPF probes are not supported on a given kernel release, so that we can avoid trying to build a config proven to be unfeasible. The goal is to save CPU and building time in the CI.

```release-note
update: avoid building unfeasible kmods and probes
```

### DIFF
--- a/cmd/root_options.go
+++ b/cmd/root_options.go
@@ -2,8 +2,10 @@ package cmd
 
 import (
 	"fmt"
+
 	"github.com/creasty/defaults"
 	"github.com/falcosecurity/driverkit/pkg/driverbuilder/builder"
+	"github.com/falcosecurity/driverkit/pkg/kernelrelease"
 	"github.com/falcosecurity/driverkit/validate"
 	"github.com/go-playground/validator/v10"
 	logger "github.com/sirupsen/logrus"
@@ -61,6 +63,14 @@ func (ro *RootOptions) Validate() []error {
 		}
 		return errArr
 	}
+
+	// check that the kernel versions supports at least one of probe and module
+	kr := kernelrelease.FromString(ro.KernelRelease)
+	kr.Architecture = kernelrelease.Architecture(ro.Architecture)
+	if !kr.SupportsModule() && !kr.SupportsProbe() {
+		return []error{fmt.Errorf("both module and probe are not supported by given options")}
+	}
+
 	return nil
 }
 
@@ -104,7 +114,7 @@ func (ro *RootOptions) toBuild() *builder.Build {
 		kernelConfigData = "bm8tZGF0YQ==" // no-data
 	}
 
-	return &builder.Build{
+	build := &builder.Build{
 		TargetType:         builder.Type(ro.Target),
 		DriverVersion:      ro.DriverVersion,
 		KernelVersion:      ro.KernelVersion,
@@ -121,6 +131,19 @@ func (ro *RootOptions) toBuild() *builder.Build {
 		RepoOrg:            ro.Repo.Org,
 		RepoName:           ro.Repo.Name,
 	}
+
+	// attempt the build in case it comes from an invalid config
+	kr := build.KernelReleaseFromBuildConfig()
+	if len(build.ModuleFilePath) > 0 && !kr.SupportsModule() {
+		build.ModuleFilePath = ""
+		logger.Warningf("Skipping build attempt of module for unsupported kernel version %s", kr.String())
+	}
+	if len(build.ProbeFilePath) > 0 && !kr.SupportsProbe() {
+		build.ProbeFilePath = ""
+		logger.Warningf("Skipping build attempt of probe for unsupported kernel version %s", kr.String())
+	}
+
+	return build
 }
 
 // RootOptionsLevelValidation validates KernelConfigData and Target at the same time.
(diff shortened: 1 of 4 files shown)


pkg/kernelrelease/kernelrelease.go
```diff
@@ -104,3 +124,11 @@ func FromString(kernelVersionStr string) KernelRelease {
 	}
 	return kv
 }
+
+func (k *KernelRelease) SupportsModule() bool {
+	return k.Compare(moduleMinKernelVersion[k.Architecture]) >= 0
```
FedeDP, line 129: `k.GTE()` :)

pkg/kernelrelease/kernelrelease.go
```diff
@@ -104,3 +124,11 @@ func FromString(kernelVersionStr string) KernelRelease {
...
+}
+
+func (k *KernelRelease) SupportsProbe() bool {
+	return k.Compare(probeMinKernelVersion[k.Architecture]) >= 0
```
FedeDP, line 133: As above! (i find it clearer to read!)

pkg/kernelrelease/kernelrelease.go
```diff
@@ -30,13 +31,32 @@ var SupportedArchs = Architectures{
...
 	for k := range SupportedArchs {
 		supportedArchsSlice[i] = k.String()
 		i++
 	}
+
+	// see compatibility matrix: https://falco.org/docs/event-sources/drivers/
+	// note: this does not make much sense for flatcar, which has a much
+	// higher major version. In that case, we assume that the module/probe
+	// is always supported, and eventually fail while building
+	moduleMinKernelVersion = make(map[Architecture]semver.Version)
```
FedeDP, line 54: Can we statically initialize this as a global, like we do for `var images` in builders.go?
I find it more readable and "fixed", to be also used as a documentation.

cmd/root_options.go
```diff
@@ -61,6 +63,14 @@ func (ro *RootOptions) Validate() []error {
 		}
 		return errArr
 	}
+
+	// check that the kernel versions supports at least one of probe and module
+	kr := kernelrelease.FromString(ro.KernelRelease)
```
FedeDP, line 68: Can we do this check later, during the `toBuild` call, while we are already doing these checks?

FedeDP (commented):
Thank you very much for this! It's really good, left some comments, but it's a tiny great PR!

cmd/root_options.go
jasondellaluce: In `toBuild` it's slightly different: assuming that the options are valid, we try to fix them if they are not optimal (e.g. one of probe or kmod is known to be unfeasible)

FedeDP: Yep i see :/ i don't know if making a new `builder.Validate()` api would help us?

FedeDP:
@dwindsor what do you think? :)

cmd/root_options.go
dwindsor: IIUC the Cmd modules individually call `rootOptions.toBuild()`, like so for the docker cmd:

```go
if !configOptions.DryRun {
    if err := driverbuilder.NewDockerBuildProcessor(viper.GetInt("timeout"), viper.GetString("proxy")).Start(rootOpts.toBuild()); err != nil {
        logger.WithError(err).Fatal("exiting")
    }
}
```

So, perhaps add `builder::Validate()`, but then call that from `cmd::toBuild()` and have `toBuild()` return `err` if it's invalid? That way we can leave existing callers as they are.

dwindsor:
Thanks for this! I imagine this feature will be liked by others who are building large amounts of drivers in-house.

jasondellaluce:
> Thanks for this! I imagine this feature will be liked by others who are building large amounts of drivers in-house. Curious: do we have any idea how many builds will be "filtered out" now?

Honestly haven't measured it! But while releasing Falco 0.33.0 the CI got stuck due to the driver building job not being able to finish in 24hrs, and the log was full of compilation failures on kernel releases we are certain it's impossible to build drivers for. I think this could lots of resources as the number of drivers to build grows.

Issif:
For the lib 3.0.1+driver, the CI built 9,084 drivers. See https://download.falco.org/driver/site/index.html?lib=3.0.1%2Bdriver&target=all&arch=all&kind=all

cmd/root_options.go
dwindsor: Sure, the original solution by Jason is fine too! =)

I'm not that opinionated about this, since after all there are only 2 callers of `toBuild()`, so updating existing users isn't a big deal.

dwindsor:
> > Thanks for this! I imagine this feature will be liked by others who are building large amounts of drivers in-house. Curious: do we have any idea how many builds will be "filtered out" now?
> 
> Honestly haven't measured it! But while releasing Falco 0.33.0 the CI got stuck due to the driver building job not being able to finish in 24hrs, and the log was full of compilation failures on kernel releases we are certain it's impossible to build drivers for. I think this could lots of resources as the number of drivers to build grows.

We went through something similar during the initial build of our driver collection. @EXONER4TED would be able to tell you better, but I believe we had to increase our build timeout twice.

This feature will eliminate a **lot** of churn.

FedeDP (approved):
/approve

PR #233, opened 2022-10-31
Improve ArchLinux Support

EXONER4TED:
Signed-off-by: Logan Bond <lbond@secureworks.com>

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area build

> /area cmd

/area pkg

> /area docs

> /area tests

Improve ArchLinux `kernelrelease` support to assume the user passes in the kernel type with the `kernelrelease`. For example: `5.10.10.zen1-1` rather than just `5.10.10-1`.

This change allows driverkit to find/build zen, lts, and hardened arch kernels - as well as the default arch stable.
Examples (which include URL output just to debug)...

Arch Stable:
```
╰─❯ _output/bin/driverkit docker --output-module /tmp/test.ko --kernelrelease 5.0.2.arch1-1 --target arch         
[https://archive.archlinux.org/packages/l/linux-headers/linux-headers-5.0.2.arch1-1-x86_64.pkg.tar.xz https://archive.archlinux.org/packages/l/linux-headers/linux-headers-5.0.2.arch1-1-x86_64.pkg.tar.zst]
INFO kernel module available                       path=/tmp/test.ko
```

Arch LTS:
```
╰─❯ _output/bin/driverkit docker --output-module /tmp/test.ko --kernelrelease 4.19.28-1 --target arch 
[https://archive.archlinux.org/packages/l/linux-lts-headers/linux-lts-headers-4.19.28-1-x86_64.pkg.tar.xz https://archive.archlinux.org/packages/l/linux-lts-headers/linux-lts-headers-4.19.28-1-x86_64.pkg.tar.zst]
INFO kernel module available                       path=/tmp/test.ko
```

Arch Zen:
```
╰─❯ _output/bin/driverkit docker --output-module /tmp/test.ko --kernelrelease 4.20.7.zen1-1 --target arch          
[https://archive.archlinux.org/packages/l/linux-zen-headers/linux-zen-headers-4.20.7.zen1-1-x86_64.pkg.tar.xz https://archive.archlinux.org/packages/l/linux-zen-headers/linux-zen-headers-4.20.7.zen1-1-x86_64.pkg.tar.zst]
INFO kernel module available                       path=/tmp/test.ko
```

Fixes # N/A

N/A

```release-note
Better support for various ArchLinux kernel types
```


EXONER4TED:
I did want to mention that I'm seeing a lot of compilation issues with various Arch kernel releases:
```
╰─❯ _output/bin/driverkit docker --output-module /tmp/test.ko --kernelrelease 5.12.8.arch1-1 --target arch -l debug
...
...
...
DEBU A+ cd /tmp/driver                     
DEBU + make CC=/usr/bin/gcc-11 KERNELDIR=/tmp/kernel 
DEBU *make -C /tmp/kernel M=/tmp/driver modules 
DEBU *make[1]: Entering directory '/tmp/kernel' 
DEBU   CC [M]  /tmp/driver/main.o         
DEBU �cc1: error: incompatible gcc/plugin versions 
DEBU cc1: error: failed to initialize plugin ./scripts/gcc-plugins/structleak_plugin.so 
DEBU Fmake[2]: *** [scripts/Makefile.build:271: /tmp/driver/main.o] Error 1 
DEBU )make[1]: Leaving directory '/tmp/kernel' 
DEBU 2make[1]: *** [Makefile:1851: /tmp/driver] Error 2 
DEBU $make: *** [Makefile:7: all] Error 2  
DEBU context canceled                             
FATA exiting                                       error="Error: No such container:path: c7b515018a108fd4376ae20d64ca90bcf4342709641c6447d139477c3c6a6e79:/tmp/driver/module.ko"
```
The same `error: failed to initialize plugin ./scripts/gcc-plugins/structleak_plugin.so` pops up in _several_ different releases. Lower level compilation issues are a bit over my head, but is this something I could potentially fix by making changes to the base `archlinux.sh` template script?

pkg/driverbuilder/builder/archlinux.go
```diff
@@ -36,59 +36,93 @@ func (c *archlinux) TemplateScript() string {
...
+						kr.Fullversion,
+						kr.FullExtraversion,
+						kr.Architecture.ToNonDeb(),
+						compressionAlgo,
+					),
+				)
+			}
+		}
+	} else if kr.Architecture.ToNonDeb() == "aarch64" {
+		baseURL := "http://tardis.tiny-vps.com/aarm/packages/l/linux-aarch64-headers/"
```
dwindsor, line 110: This is an odd mirror

EXONER4TED: I agree! See here: https://github.com/falcosecurity/kernel-crawler/pull/61#issuecomment-1297220299

Neither Federico or I could find a better spot for `aarch64` hosted Arch kernel headers... Federico had been using this one previously, so I kept it on the off chance someone is using driverkit with `aarch64`.

EXONER4TED:
I was able to fix the gcc-12 issue by using `--builderimage` with a custom archlinux container:
```
FROM archlinux:latest

RUN pacman -Sy && pacman -Sy --noconfirm make pahole gcc11 gcc11-libs gcc && ln -s /usr/bin/gcc /usr/bin/gcc-12
```
...and then it works. Kinda makes sense given how new Arch is - `--builderimage` may just be required for Arch using newer tools than Debian.

FedeDP:
So, fact is: we need to use some custom builderimage based on top of arch to build some arch kernels, right? I think they might use some gcc plugin that is not supported by the debian one.
I think it is good to go for now; we don't aim at natively supporting everything; good thing is one can build its own builder image and build all the archlinux kernels if needed.

> DEBU (Failed to parse base BTF 'vmlinux': -22 

Note that this does not depend upon gcc12, but it seems to be an issue with 6.x kernels; libs CI has a job that tests the build with driverkit on each maj kernel version (2.x, 3.x, 4.x, 5.19.12 on arch), and we are not seeing any issue.
I am able to build 5.19.12 with driverkit on my pc too.

After some research with the help of @Andreagit97 , we found out that newer kernels are enforcing the generation of BTF for kmod too.  
Unfortunately, this is somewhat harming kmod build portability, as it seems we are not able to build a kmod for a distro different than the one that is running anymore.
Some links:
* https://groups.google.com/g/linux.debian.kernel/c/2ePnmlEXeIQ/m/mdindWfVEAAJ
* https://lore.kernel.org/bpf/20201113103158.GA30836@linux-8ccs/t/
* https://groups.google.com/g/linux.debian.bugs.dist/c/Wv6lQZycgHY

The only workaround that we found, right now, is to forcefully disable the Makefile step that generates BTF, that is:
```
quiet_cmd_btf_ko = BTF [M] $@
cmd_btf_ko =                                                      \
if [ -f vmlinux ]; then                                         \
    LLVM_OBJCOPY="$(OBJCOPY)" $(PAHOLE) -J $(PAHOLE_FLAGS) --btf_base vmlinux $@; \
    $(RESOLVE_BTFIDS) -b vmlinux $@;                        \
else                                                            \
   printf "Skipping BTF generation for %s due to unavailability of vmlinux\n" $@ 1>&2; \
fi;
```

Basically, we'd need to rename `/lib/modules/$(uname -r)/build/vmlinux` to something else, and at the end of the kmod build, rename it back.
In this way, the above piece of code, will just print "Skipping BTF generation...".
WDYT?

FedeDP:
Btw i think we can tackle it in another PR, i can work on that if we agree on the solution.

dwindsor:
> Unfortunately, this is somewhat harming kmod build portability, as it seems we are not able to build a kmod for a distro different than the one that is running anymore.

Ugh, this is not great.

Maybe a bit late to the party here, but why is BTF type info necessary for building a kmod?

FedeDP:
cc @Andreagit97 ; i am not sure why, but upstream is requiring this now!

EXONER4TED:
Hm.. That is interesting. I guess my question would be how does ArchLinux container build it just fine and the Debian one not? It doesn't seem to care about the `vmlinux`:
```
╰─❯ _output/bin/driverkit docker --output-module /tmp/test.ko --kernelrelease 6.0.6.arch1-1 --target arch -l debug --builderimage arch-test:latest
...
...
...
DEBU  + modinfo /tmp/driver/module.ko      
DEBU �filename:       /tmp/driver/module.ko 
DEBU schema_version: 2.2.0                        
DEBU api_version:    2.0.0                        
DEBU build_commit:   master                       
DEBU version:        master                       
DEBU author:         the Falco authors            
DEBU license:        GPL                          
DEBU srcversion:     3BDA4BA7B88A29C8A2E6F49      
DEBU depends:                                     
DEBU retpoline:      Y                            
DEBU name:           falco                        
DEBU vermagic:       6.0.6-arch1-1 SMP preempt mod_unload  
DEBU parm:           g_buffer_bytes_dim:This is the dimension of a single per-CPU buffer in bytes. Please note: this buffer will be mapped twice in the process virtual memory, so pay attention to its size. 
DEBU parm:           max_consumers:Maximum number of consumers that can simultaneously open the devices (uint) 
DEBU parm:           verbose:Enable verbose logging (bool) 
INFO kernel module available                       path=/tmp/test.ko
```

That being said, I'm good with adding the `mv vmlinux` to somewhere and putting it back. We could just add that to the template script inside the `{{ if .BuildModule }}` section even...

Otherwise, I added the archlinux container to the `build/` folder just as docs, but we are good to build ArchLinux in that way for now if we don't want to make any changes to the Makefile or scripts right away.

dwindsor:
> That being said, I'm good with adding the mv vmlinux to somewhere and putting it back. We could just add that to the template script inside the {{ if .BuildModule }} section even...

Well, we'll now have to do this in each builder IIUC, so might as well just do it in one PR as @FedeDP suggested,

Andreagit97:
Agree with @FedeDP @dwindsor we need to investigate more on that before taking a decision

EXONER4TED:
Makes sense to me! Then I suppose this PR is ready to go? I can keep the Arch Dockerfile in the repo for the sake of documentation for anyone who might be trying this besides us.

FedeDP:
I'd avoid keeping the builder arch because otherwise a new builder image will be pushed; perhaps a note under the examples for archlinux is enough for now: https://github.com/falcosecurity/driverkit/blob/master/Example_configs.md#archlinux

We will tackle it later ;)

EXONER4TED:
Ah I see.. there's just a loop in the Makefile to build anything there. Gotcha, I'll remove it and update documentation.

dwindsor (approved):

PR #235, opened 2022-11-04
feat: add development fedora repos to enable building of driver for fc37 which isn't in final release yet

abroglesc:
**What type of PR is this?**

/kind feature

/area pkg

Previously when attempting to build the driver for `5.19.16-301.fc37.x86_64_1` we could not. This is because the fedora core 37 isn't in final release and is still in the development section of the repo so it couldn't actually find and download the kernel headers. By adding this, we are now able to download the kernel headers and ultimately build fc37 probes.


abroglesc:
Tagging @EXONER4TED as he created initial fedora support. Thank you for that btw super helpful!

EXONER4TED (approved):

PR #238, opened 2022-12-09
feat(builder-image): config flag for custom base image

Lowaiz:
config flag for custom base image to override the compile-time defined BaseImage

Signed-off-by: Lyonel Martinez <lyonel.martinez@numberly.com>

**What type of PR is this?**

/kind feature

 /area cmd

 /area pkg

 /area docs

In order to use proxy images and to avoid defining a single image to the command line and still benefit the auto-choose based on GCC version need by the kernel

None

I included a mini fix of the Makefile, the `make docs` was erasing builder and builder_image doc.
And it seems that a pretty big part of the docs was missing.

```release-note
new option: --builderimagebase base image string used to get the different build images 
```

### DIFF
--- a/cmd/root_options.go
+++ b/cmd/root_options.go
@@ -33,6 +33,7 @@ type RootOptions struct {
 	Target           string   `validate:"required,target" name:"target"`
 	KernelConfigData string   `validate:"omitempty,base64" name:"kernel config data"` // fixme > tag "name" does not seem to work when used at struct level, but works when used at inner level
 	BuilderImage     string   `validate:"omitempty,imagename" name:"builder image"`
+	BuilderImageBase string   `validate:"omitempty,imagename" name:"builder base image"`
 	GCCVersion       string   `validate:"omitempty,semvertolerant" name:"gcc version"`
 	KernelUrls       []string `name:"kernel header urls"`
 	Repo             RepoOptions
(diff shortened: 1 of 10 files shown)


FedeDP:
Hi @Lowaiz ! Thanks for this PR
> In order to use proxy images and to avoid defining a single image to the command line and still benefit the auto-choose based on GCC version need by the kernel

If i understood correctly, you want to achieve a situation where you can ship your own `_bookwork/_bullseye/_buster` images, right? 
Can you expand on the use cases?

I actually think it would be great to fetch tags (like `DRIVERKIT_PROVIDED_GCCS`) `docker inspect`ing the provided images, allowing users to fully customize the list of builder images and entirely dropping our static map.
BUT we cannot afford to pull all the builder images before every build; it would increase build time 10x... that's why i used the static map.
Yet it would be so freakingly good to be able to fetch metadata from provided builder images...

Lowaiz:
> If i understood correctly, you want to achieve a situation where you can ship your own _bookwork/_bullseye/_buster images, right?
Can you expand on the use cases?

In order to limit requests to docker.io from our infra, we are proxying images by pushing them in our private container registry. So with the Driverkit, we need to push the `driverkit_*` images to our registry, but as the image link is hard coded we couldn't use these images. That why we need the new flag.
So it's not really about shipping our own custom images as we don't modify them when we add them in the registry.

> BUT we cannot afford to pull all the builder images before every build; it would increase build time 10x... that's why i used the static map

The static map is not problematic for us. As I said, the proxyed images are not modified as we want to stick to the official behavior of Driverkit. It might be usefull to add a bit of doc to explain that this new flag assume that you have no custom builder images but only proxy images.

FedeDP:
Oh i see, it makes sense! Thank you! I will give this another look asap!

cmd/root_options.go
```diff
@@ -33,6 +33,7 @@ type RootOptions struct {
 	Target           string   `validate:"required,target" name:"target"`
 	KernelConfigData string   `validate:"omitempty,base64" name:"kernel config data"` // fixme > tag "name" does not seem to work when used at struct level, but works when used at inner level
 	BuilderImage     string   `validate:"omitempty,imagename" name:"builder image"`
+	BuilderImageBase string   `validate:"omitempty,imagenamebase" name:"builder base image"`
```
FedeDP, line 36: I think we can just reuse `imagename` validate tag here, without creating a new one!

FedeDP (commented):
Left a small comment, otherwise this LGTM!

cmd/root_options.go
Lowaiz: Applied

FedeDP (approved):
/approve

FedeDP:
Mmmh poiana?
Trying to close and reopen to trigger.

Issue #236, opened 2022-11-18
Cannot Build CentOS 7: `3.10.0-1160.80.1.el7.x86_64`


EXONER4TED:
**Describe the bug**

It looks like CentOS 7 got a new kernel ~Nov. 10 - at least, that's when mirrors picked it up. Our build pipelines internally have been failing for ALL falco libs versions, pointing to the problem being something else. Trying to manually build it gives an error.

Command:
```
 _output/bin/driverkit docker --output-module test.ko --kernelrelease 3.10.0-1160.80.1.el7.x86_64 --kernelversion 1 --driverversion 3.0.0+driver --target centos --moduledevicename scwx-falco --moduledrivername scwx-falco --kernelurls https://mirrors.edge.kernel.org/centos/7/updates/x86_64/Packages/kernel-devel-3.10.0-1160.80.1.el7.x86_64.rpm -l debug
```
Error:
```
DEBU !  CC [M]  /tmp/driver/tp_table.o     
DEBU #  LD [M]  /tmp/driver/scwx-falco.o   
DEBU ld: Q/tmp/driver/tp_table.o: invalid string offset 1819042147 >= 21 for section `mtab' 
DEBU /tmp/driver/tp_table.o3: error adding symbols: File format not recognized 
DEBU Lmake[2]: *** [scripts/Makefile.build:470: /tmp/driver/scwx-falco.o] Error 1 
DEBU )make[1]: Leaving directory '/tmp/kernel' 
DEBU :make[1]: *** [Makefile:1316: _module_/tmp/driver] Error 2 
DEBU $make: *** [Makefile:7: all] Error 2  
DEBU log pipe close                                error=EOF
```
It _looks_ like it's an issue with the linker..?

Either way to get around this problem, I tried creating a custom builder image using CentOS 7 as the base:
```
FROM centos:7

RUN yum install -y make which binutils gcc && ln -s $(which gcc) /usr/bin/gcc-4.9
```
And using `--builderimage` with this image works just fine 🙂 

**NOTE:** GCC in that above image is actually gcc-4.8.5, NOT gcc-4.9 - I just symlink the `/usr/bin/gcc` to `/usr/bin/gcc-4.9` to trick the driverkit GCC selector into using my 4.8.5 - and that works! It gets more interesting... If I change the driverkit GCC selector to force gcc-4.8 for CentOS 3.10 kernels, that also fails... It seems that specifically for this NEW CentOS kernel, it requires gcc-4.8.5? 

[...]

PR #244, opened 2022-12-22
new: dynamic builders

### LINKED ISSUE #241: Improve build image selection
**Motivation**

I don't really like the static map for builder images.
It was a quick way to reach multiple-builders support, but is far from best approach.

**Feature**

Instead of using a static map, do:

* `docker search falcosecurity/driverkit` to fetch list of builder images
* their name should be `builder-$target-$gccX_$gccY` so that we can extract every needed info from that
* This also allows to push multiple images for each target
* if `$target` is not present, fallback to `any` target
* same algorithm we already got to choose the image

One can then specify a different docker repo to fetch from (fixing also #238) or force-set a builder image, just like we support now.

**Alternatives**

We could also use docker image labels to set supported GCCs, but that would require `inspect`ing each image, thus downloading them, greatly increasing startup time.

**Additional context**

With such a feature in place, we might start to add new builder images for whenever a target needs it, fixing issues such as #236.
Let's say that a centOS build needs gcc 5.1; 
* `docker search` will tell us that we ship a gcc-4.9 `any` image;
* but then we add a centos_gcc5.1 image
* so, given that `$target` would match with our build (ie: we are building centos, so we consider all `any` builder images, + the `centos` ones), we can now choose from the new builder image too
* automatic algorithm will then choose the best-fit image (ie: the new 5.1 one)
* we build the driver

So, adding a new builder image would just be a matter of adding its dockerfile, and updating the default selection algorithm for the specific case, if needed.
I think the best we could do is to also improve `GCCVersion(kr kernelrelease.KernelRelease) semver.Version` interface to let it switch to default algorithm if eg: empty `semver.Version` is returned; this way, we could let any target implement target-specific gcc version tricks, but the default cases would still remain the same for all of them.

**Only big impact i see** is for CI, given that we might need to build lots of builder images.
Currently, after lots of small improvements, CI takes ~15min, so i think that adding new builder images is feasible (i think we can add at least 4-5 more).
Therefore, we might avoid thinking about CI times for now.

(i don't know, perhaps we might cache images somehow too? I mean, most of the times the builder images are untouched, but the master CI has to rebuild them and push them just to update the image tag...)

/cc @dwindsor @EXONER4TED @Lowaiz

FedeDP:
**What type of PR is this?**

/kind feature

/area build
/area cmd
/area pkg

This PR implements the proposed improvements to builder images: #241 .

Basically, we do now allow to pass a list of docker repos in descending priority order.
`falcosecurity/driverkit` is always emplaced back, so that is has lowest available priority.

List of images is then dynamically and transparently loaded at runtime, given that images in the passed repositories follow this scheme for their name: `driverkit-builder-(?P<target>[a-z0-9]+)-(?P<arch>x86_64|aarch64)(?P<gccVers>(_gcc[0-9]+.[0-9]+.[0-9]+)+)$`, basically, it means something like this: `myrepo/driverkit-builder-ubuntu-x86_64_gcc5.8.0_gcc7.0.0`.

NOTE: provided images should expose full gcc semver fields (ie: even 0 ones), like the above example.
Internally, they should soft link gcc to the full semversioned name, like: `ln -s /usr/bin/gcc-5 /usr/bin/gcc-5.0.0`.
IMHO it should not be a big issue. Let me know wyt! We can try to find another solution if you think it is too time consuming doing that (but again, i don't really think so).

This PR should also improve the situation for @Lowaiz and its #238 patch.

"any" target builder images, are fallback images to be used when a specific-target one is missing. The algorithm goes as follow:
* get required target GCC from either:
* * command line argument (`--gccversion`) again, as a full semver string
* * GCCVersionRequestor interface by the builder
* * not-so-smart algorithm given a kernelrelease
* then, load all available images from al specified docker repos
* for each discovered image, create an entry for any of the provided gcc in a map whose key is `$target_$gccversion`
* finally, try to find an image that provides target GCC, by first looking to `$target_$gccversion`, then, if it could not be found, looking at `any_$gccversion`
* if any image is found, we are done
* else, loop on all images that are either for target $target, or "any", to find the one that provides the nearest gcc version
* we are over ;) 
* if you ask for a specific `--builderimage`, you **must** provide the requested GCC (or force a `--gccversion`); this was already enforced

Fixes #241 

```release-note
new: added support for dynamic builder images
```


FedeDP:
Note: CI will fail badly of course, because images cannot be found.

Moreover, i still need to update multiple docs files.

FedeDP:
Example test (i pushed an image to my own dockerhub):
```
./_output/bin/driverkit docker --kernelrelease 6.0.12.arch1-1 --target arch --output-module /tmp/arch.ko  --output-probe /tmp/arch.o -l debug --dockerrepo docker.io/fededp/driverkit
```

FedeDP:
Possible way to fix tests:
* push all default images to eg: fededp dockerhub
* update tests makefile to add `--dockerrepo docker.io/fededp/driverkit`
* merge this
* open a new PR to drop the makefile trick
* we are done

FedeDP:
As expected, integration tests are failing with:
```
_output/bin/driverkit docker -c test/x86_64/configs/al.yaml --builderimage auto:master -l debug --timeout 600
INFO using config file                             file=test/x86_64/configs/al.yaml
DEBU running with options                          arch=amd64 driverversion=master kernelrelease=4.14.128-87.105.amzn1.x86_64 kernelurls="[https://download.falco.org/fixtures/driverkit/x86_64_devel/kernel-devel-4.14.128-87.105.amzn1.x86_64.rpm]" kernelversion=1 output-module=/tmp/amazonlinux_x86_64.ko repo-name=libs repo-org=falcosecurity target=amazonlinux
DEBU doing a new docker build                     
DEBU kernel header url found                       url="https://download.falco.org/fixtures/driverkit/x86_64_devel/kernel-devel-4.14.128-87.105.amzn1.x86_64.rpm"
DEBU Malformed image name: falcosecurity/driverkit-builder 
DEBU Malformed image name: falcosecurity/driverkit 
DEBU Malformed image name: falcosecurity/driverkit-builder_bookworm 
DEBU Malformed image name: falcosecurity/driverkit-builder_buster 
DEBU Malformed image name: falcosecurity/driverkit-builder_bullseye 
2022/12/22 16:23:07 Could not load any builder image. Leaving.
```

FedeDP:
Instead of naming the builder images using the aforementioned regex, we could build a solution that uses http docker registry API to fetch image labels for each image without actually pulling the image; this is done eg by this project: https://github.com/felicianotech/sonar/blob/7d98bae716997a3016c74ec5b1518601e5ca59ae/sonar/docker/label.go#L11
and explained on a medium article: https://medium.com/hackernoon/inspecting-docker-images-without-pulling-them-4de53d34a604

Let's assume this is in place, we could have multiple labels:
* DriverkitArch:...
* DriverkitGCC:4.8.0,5.0.0,7.0.0
* DriverkitTarget: ...

In the end, this method allows for cleaner builder images dockerfiles; moreover has the big plus that LABELS and installed GCCs are declared in the very same context (ie: the dockerfile), therefore it is harder to have it wrong (while is much easier to get image name wrong instead).

At the same time, i am afraid that the http api could change  (eg: being bumped to v3) and therefore we would need more work to keep our code up to date.

WDYT?

FedeDP:
On a second thought, the problem with the http registry API version is performance: imagine we need to scrape up to 100 builder images for each repo; this means we are doing 100 http requests for each repo listed; assuming a 10ms ping, we are losing 1s for the build for each repo listed.
Driverkit must be as quick as possible IMHO.

FedeDP:
The new images command output example:
```
./_output/bin/driverkit images --kernelrelease 6.0.12.arch1-1 --target arch --output-module /tmp/arch.ko  --output-probe /tmp/arch.o -l debug --dockerrepo docker.io/fededp/driverkit
DEBU running without a configuration file
DEBU running with options                          arch=amd64 driverversion=master kernelrelease=6.0.12.arch1-1 kernelversion=1 output-module=/tmp/arch.ko output-probe=/tmp/arch.o repo-name=libs repo-org=falcosecurity target=arch
INFO listing images                                processor=images
|                         IMAGE                          | TARGET | ARCH  |  GCC   |
|--------------------------------------------------------|--------|-------|--------|
| fededp/driverkit-builder-any-x86_64_gcc10.0.0_gcc9.0.0 | any    | amd64 | 10.0.0 |
| fededp/driverkit-builder-any-x86_64_gcc10.0.0_gcc9.0.0 | any    | amd64 | 9.0.0  |
```

FedeDP:
TODO:
- [x] update doc
- [ ] provide an image that fixes #236 
- [ ] Test ...??

dwindsor (dismissed):
> NOTE: provided images should expose full gcc semver fields (ie: even 0 ones), like the above example.
Internally, they should soft link gcc to the full semversioned name, like: ln -s /usr/bin/gcc-5 /usr/bin/gcc-5.0.0.
IMHO it should naot be a big issue. Let me know wyt! We can try to find another solution if you think it is too time consuming doing that (but again, i don't really think so).

Agreed, adopting this convention of linking to a semver gcc is not a big deal. Thanks for this feature! In terms of maintenance, we don't have to maintain the static map of gcc versions anymore correct?

FedeDP:
> Agreed, adopting this convention of linking to a semver gcc is not a big deal. Thanks for this feature! In terms of maintenance, we don't have to maintain the static map of gcc versions anymore correct?

Exactly! 
I wish to at least provide a new image for centos to fix #236 ; can you help me? I tried with a centos one, but it is not working :( 
Then, i'd need to fix tests before eventually merging this!

FedeDP:
Rebased on top of master.
Still looking forward to provide a new builder image to fix #236 :) then we can try to find a way to fix CI and eventually merge this, if we gather enough positive feedback!
@Lowaiz @EXONER4TED @dwindsor

FedeDP:
Ok, i found the issue with my centos7 gcc 4.8.5 builder image; it is now able to build centos `3.10.x` drivers.

FedeDP:
To fix CI we might want to use my `fededp` repo for a run, and then switch back to use `falcosecurity` repo in a subsequent PR.
But building the 3 arm64 builder images through qemu will take hours on my x86 laptop :) 
WDYT?

FedeDP:
>  But building the 3 arm64 builder images through qemu will take hours on my x86 laptop :)

I used an arm ec2 instance to build the arm images.
Latest commit enforce the usage of `docker.io/fededp/driverkit` dockerrepo during integration tests; we should now be able to actually test them.

FedeDP:
After latest commits, i fixed `--gccversion` support in both `images` cmd and build.

FedeDP:
IMHO the PR is ready; removed the wip; adding the hold label like for a week or so, to wait to gather some more feedback :)
/hold

dwindsor:
Thank you for the incredible amount of work that's gone into this. 🙏

How do you feel about changing the name of the new driverkit command-line arg from `--dockerrepo` to something like` --builderrepo`, so as to indicate to the user that said repo is for hosting the image in `--builderimage`?

FedeDP:
Uh it is a neat idea! Thank you!

FedeDP:
For a future CLI api break...i feel like `builder.repo builder.image` would be a much better naming (ie: having a new `builder.` config group)?

The same could apply to `repo-org` and `repo-name` too!

Like:
```
A command line tool to build Falco kernel modules and eBPF probes.

Usage:
driverkit
driverkit [command]

Available Commands:
completion            Generates completion scripts.
docker                Build Falco kernel modules and eBPF probes against a docker daemon.
help                  Help about any command
images                List builder images
kubernetes            Build Falco kernel modules and eBPF probes against a Kubernetes cluster.
kubernetes-in-cluster Build Falco kernel modules and eBPF probes against a Kubernetes cluster inside a Kubernetes cluster.

Flags:
--architecture string       target architecture for the built driver, one of [amd64,arm64] (default "amd64")
-c, --config string             config file path (default $HOME/.driverkit.yaml if exists)
--driverversion string      driver version as a git commit hash or as a git tag (default "master")
--dryrun                    do not actually perform the action
--gccversion string         enforce a specific gcc version for the build
-h, --help                      help for driverkit
--kernelconfigdata string   base64 encoded kernel config data: in some systems it can be found under the /boot directory, in other it is gzip compressed under /proc
--kernelrelease string      kernel release to build the module for, it can be found by executing 'uname -v'
--kernelurls strings        list of kernel header urls (e.g. --kernelurls <URL1> --kernelurls <URL2> --kernelurls "<URL3>,<URL4>")
--kernelversion string      kernel version to build the module for, it's the numeric value after the hash when you execute 'uname -v' (default "1")
-l, --loglevel string           log level (default "info")
--moduledevicename string   kernel module device name (the default is falco, so the device will be under /dev/falco*) (default "falco")
--moduledrivername string   kernel module driver name, i.e. the name you see when you check installed modules via lsmod (default "falco")
--name string               repository github name (default "libs")
--org string                repository github organization (default "falcosecurity")
--output-module string      filepath where to save the resulting kernel module
--output-probe string       filepath where to save the resulting eBPF probe
--proxy string              the proxy to use to download data
-t, --target string             the system to target the build for, one of [almalinux,amazonlinux,amazonlinux2,amazonlinux2022,arch,bottlerocket,centos,debian,fedora,flatcar,minikube,opensuse,photon,redhat,rocky,ubuntu,vanilla]
--timeout int               timeout in seconds (default 120)
-v, --version                   version for driverkit

Builders:
--image string       docker image to be used to build the kernel module and eBPF probe. If not provided, an automatically selected image will be used.
--repo strings        list of docker repositories in descending priority order, used to search for builder images. Default falcosecurity/driverkit will always be enforced as lowest priority repo. eg: --dockerrepo myorg/driverkit --dockerrepo falcosecurity/driverkit
```

And so on, even `kernel{urls,version,release,configdata}` could take advantage of this.
It would help the user better understand various driverkit option.

Cobra does not support this right now but a PR might get merged soonish: https://github.com/spf13/cobra/pull/1778

FedeDP:
Done @dwindsor  :)

FedeDP:
Run `make docs` too! Forgot about it!

dwindsor (dismissed):

FedeDP:
Improved documentation! :page_facing_up:

FedeDP:
@dwindsor can you reapprove since i pushed some more docs? :)

dwindsor (approved):
Goodbye static list of builders =)

PR #257, opened 2023-03-17
fix(pkg): fix gcc version from builder images setter.

FedeDP:
**What type of PR is this?**

/kind bug

> /area pkg

Skip all the fallback logic if we find an image providing the correct gcc in the first place.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/driverbuilder/builder/builders.go
+++ b/pkg/driverbuilder/builder/builders.go
@@ -185,31 +185,31 @@ func (b *Build) setGCCVersion(builder Builder, kr kernelrelease.KernelRelease) {
 	image, ok := b.Images.findImage(b.TargetType, targetGCC)
 	if ok {
 		b.GCCVersion = image.GCCVersion.String()
-	}
-
-	// Step 2:
-	// Build the list of "proposed" GCC versions,
-	// that is, the list of available gccs from images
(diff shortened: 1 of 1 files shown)


pkg/driverbuilder/builder/builders.go
```diff
@@ -185,31 +185,31 @@ func (b *Build) setGCCVersion(builder Builder, kr kernelrelease.KernelRelease) {
...
-	// that is, the list of available gccs from images
-	// for each builder image
-	proposedGCCs := make([]semver.Version, 0)
-	for _, img := range b.Images {
-		proposedGCCs = append(proposedGCCs, img.GCCVersion)
-		logger.WithField("image", img.Name).
-			WithField("targetGCC", targetGCC.String()).
-			Debug("proposedGCC=", img.GCCVersion.String())
-	}
+	} else {
```
FedeDP, line 188: If we found a `gccversion` provided by an image, there is no need to fallback at secondary algorithm.

EXONER4TED:
/approve

One more try 😅

EXONER4TED:
Nope, it wants someone other than me 🤷 Looks good to me though :)

EXONER4TED (approved):

Issue #256, opened 2023-03-14
Usage and break of builder image listing


Lowaiz:
**Motivation**
The new feature of auto listing of builder image totally break the usage of Driverkit inside a deployed Kubernetes Pod.

The function ImageSearch seems to look for the local unix socket, but, as the DinD is an unsafe pratice, this kind of right (accessing Docker socket, so Privileged pods) is not an option I want to deploy. 
Resulting ImageSearch to fail, causing the build to fail. 

Another problem with the current implementation is the authentication to private repositories that seems not handle. And it seems that the command `docker search` struggle to work with gitlab container registry,

In another topic, I find putting gcc version inside the builder image name not quite optimum. A solution could be to apply multiple tag to images containing multiple gcc version, i.e.:
`driverkit-builder-any-x86_64:gcc-8.0.0`, `driverkit-builder-any-x86_64:gcc-6.0.0`, `driverkit-builder-any-x86_64:gcc-5.0.0`, ... could point the same image containing the multiple gcc versions avoiding to create ultra long name, creating a better flavoring.

**Feature**

- Fix the usage of Docker SDK Client to be able to search without the Docker unix socket + token/auth management for custom repos 
- Possible rename/rework of builder images organization

**Alternatives**

I tried to deep dive the docker SDK, but I can't find a way to fix it.
The usage of Docker search API function add a big complexity while a simple check of present of base image name, with a possible prefix, and with flavoured tag for gcc and with a possible flavoured tagname for target-specific build (other images than `any`ones) could erase most of that complexity.

FedeDP:
Hi @Lowaiz! 
Sorry for breaking kubernetes :/ We left the issue and the PR opened for like 2 months because i knew something could go wrong. I also pinged you directly in the PR because i know you are the one using kubernetes; that's unfortunate!

> The function ImageSearch seems to look for the local unix socket, but, as the DinD is an unsafe pratice, this kind of right (accessing Docker socket, so Privileged pods) is not an option I want to deploy.
Resulting ImageSearch to fail, causing the build to fail.

Yep i agree, didn't think of this aspect since we just use the docker builder. I think we can easily fix this by providing a `--builderindex` option that overrides the search logic with a fixed list of images. 
It would make a lot of sense too IMHO.

> Another problem with the current implementation is the authentication to private repositories that seems not handle. And it seems that the command docker search struggle to work with gitlab container registry,

About gitlab container registry, that's weird :/ 
[...]

PR #258, opened 2023-03-17
feat(builder-indexes): cli flag to build Images map 

Lowaiz:
feat(builder-indexes): cli flag to build Images map without scanning docker registries

**What type of PR is this?**

/kind feature

/area cmd

/area pkg

/area docs

New flag to be able to set the builder.Images map with data given via cli to not search and parse Docker registries.
The flag `builderindex` is string slice, that I parse by splitting with `;` character, considering the first term as the image name, the second term as the target, and everything else as GCC versions available on the image.

I guess the format is not optimum, but I can't find other and simpler way to parse this kind of info from CLI:

Usage: --builderindex "`<image>`;`<target>`;`<gcc-version>`;`<gcc-version>`;`...`" 
Example: --builderindex "`falcosecurity/driverkit-builder`;`any`;`4.9.0`;`5.0.0`;`6.0.0`;`8.0.0`"

Multiple --buildindex can be set to insert different images

None, context here: #256

```release-note
- New CLI flag '--builderindex' making able to set up the Images map from CLI.
  Usage: --builderindex "<image>;<target>;<gcc-versuin>;<gcc-versuin>;...".
  Example: --builderindex "falcosecurity/driverkit-builder;any;4.9.0;5.0.0;6.0.0;8.0.0"
```


FedeDP:
Thanks for this PR!
I was wondering whether to use the same `builderrepo` option instead; like:
* if `builderrepo` points to a local file, we then expect it to provide a list of `"<image>;<target>;<gcc-version>,<gcc-version>,<gcc-version>\n`
* else, we expect it to be a docker repo

It is simpler; what i don't really like about having the new option (thinking twice about it) is how it plays together with `builderrepo` for example, given that they both provide the same informations.

Moreover, it would be huge to abstract everything behind an interface-like object, eg:
```go
type ImagesLister interface {
   ListImage() []Images
}

type FileImagesLister struct {
  *os.File
}

func (f *FileImagesLister) ListImage() []Images { ... }

type RepoImagesLister struct {
  repo string
}

func (f *RepoImagesLister) ListImage() []Images { ... }

RootOptions.go:
for builderrepo := range builderrepos {
  if isFile(builderrepo) {
     builderepostlister.append(newFileImagesLister(builderrepo))
 else
    builderepostlister.append(newRepoImagesLister(builderrepo))
}
```
Then basically, `images.findImage()` will receive the list of `builderepostlister`(that will actually be stored in `builder.Build`), and calls `ListImage()` on each one. 
This way, we also support using `image` subcmd with the new feat.

WDYT?
I know it is a bigger change but it will be much better IMHO!

Lowaiz:
Hey,

That's a clever idea. And when changing to falctoctl later will be simpler.

pkg/driverbuilder/builder/image.go
```diff
@@ -19,9 +19,24 @@ type Image struct {
...
+
+type FileImagesLister struct {
+	file     *os.File
+	Build    *Build
+	FilePath string
+}
+
+type RepoImagesLister struct {
+	Repo  string
+	Build *Build
```
FedeDP, line 34: I'd avoid storing the `Build` here! Instead, i'd pass it to the `ListImages` methods, as a param.
I think it is better for these methods to be stateless: just return a list of `Image`s.
Moreover, given that `regs` are now just used by `RepoImagesLister`, i think they can just be declared inside the `ListImages` method.

pkg/driverbuilder/builder/image.go
```diff
@@ -19,9 +19,24 @@ type Image struct {
...
+}
+
 type ImageKey string
 
-func (i *Image) toKey() ImageKey {
+func (i *Image) ToKey() ImageKey {
```
FedeDP, line 39: Why are you exposing this method now?

pkg/driverbuilder/builder/builders.go
```diff
@@ -140,7 +140,7 @@ func defaultGCC(kr kernelrelease.KernelRelease) semver.Version {
 	}
 }
 
-func mustParseTolerant(gccStr string) semver.Version {
+func MustParseTolerant(gccStr string) semver.Version {
```
FedeDP, line 143: Why are you exposing this method now?

pkg/driverbuilder/builder/image.go
FedeDP: I think they can also be global const variables in image.go, and then referenced in each `RepoImagesLister`; i don't like adding yet another non-build related field to `Build` struct.
Given it is only used by `RepoImagesLister`, i think it should live there!

pkg/driverbuilder/builder/image.go
```diff
@@ -45,78 +63,110 @@ func (im ImagesMap) findImage(target Type, gccVers semver.Version) (Image, bool)
...
+		name := infos[0]
+		target := Type(infos[1])
+		gccVersions := infos[2:]
+		for _, gcc := range gccVersions {
+			buildImage := Image{
+				Name:       name,
+				Target:     target,
+				GCCVersion: mustParseTolerant(gcc),
+			}
+			build.Images[buildImage.toKey()] = buildImage
```
FedeDP, line 87: I'd just make this method return a slice of Images, it is a much better API (and is unit testable); then main `image.ListImages` will gather all the images and fill the build.Images map.

Lowaiz: Found a way to eliminate ref to builder inside objects and function of ImageLister

pkg/driverbuilder/builder/image.go
```diff
@@ -45,78 +63,123 @@ func (im ImagesMap) findImage(target Type, gccVers semver.Version) (Image, bool)
...
+			// and we cannot guarantee here that any subsequent docker repos
+			// does not provide a target-specific image that offers same gcc version
+			for _, gccVer := range gccVers {
+				// If user set a fixed gcc version, only load images that provide it.
+				buildImage := Image{
+					GCCVersion: mustParseTolerant(gccVer),
+					Name:       img.Name,
+				}
+				if regIdx == 0 {
+					buildImage.Target = Type("target-placeholder")
```
FedeDP, line 159: I think the best here would be to properly capture the type in the regexes, so that we can just do `buildImage.Target = Type(regType)`
be it `any` or a specific target.

FedeDP: And we can then skip the `target-placeholder` trick below.

pkg/driverbuilder/builder/image.go
```diff
@@ -45,78 +63,123 @@ func (im ImagesMap) findImage(target Type, gccVers semver.Version) (Image, bool)
...
+				} else {
+					buildImage.Target = Type("any")
 				}
+				res = append(res, buildImage)
 			}
 		}
 	}
-	if len(b.Images) == 0 {
```
FedeDP, line 119: I'd keep this check, to make sure that we loaded at least an image.

cmd/root_options.go
```diff
@@ -132,11 +132,21 @@ func (ro *RootOptions) toBuild() *builder.Build {
 		KernelUrls:       ro.KernelUrls,
 		RepoOrg:          ro.Repo.Org,
 		RepoName:         ro.Repo.Name,
+		Images:           make(builder.ImagesMap),
+	}
+
+	if len(build.BuilderRepos) == 0 {
```
FedeDP, line 138: I'd inizialize it as a default value during declaration, using the `default:` tag; we use this library: https://github.com/creasty/defaults#usage

As you can see, a 
```
BuilderRepos []string `default: ["docker.io/falcosecurity/driverkit"] validate:"omitempty" name:"docker repositories to look for builder images or absolute path pointing to a file container builder image index"`
```
Is enough.

FedeDP:
Last round of review, i think after these ones will be addressed, we will be ready :rocket:

FedeDP:
Btw this is already good to go; if you are willing to spend a little bit more time on refining this, it's great! Otherwise i can do that myself on top of yours once it's merged :)
Sorry for the long list of "nit picks", but i feel like we are doing a really good future-proof job here!

cmd/root_options.go
Lowaiz: Oh I didn't know we could place complex types inside this default statement

Lowaiz:
@FedeDP 
I guess, it's good now !

FedeDP:
Yep @Lowaiz!
Thanks for your patience! It LGTM! :)

cmd/root.go
```diff
@@ -144,7 +144,7 @@ func NewRootCmd() *RootCmd {
...
 	flags.StringVar(&rootOpts.BuilderImage, "builderimage", rootOpts.BuilderImage, "docker image to be used to build the kernel module and eBPF probe. If not provided, an automatically selected image will be used.")
-	flags.StringSliceVar(&rootOpts.BuilderRepos, "builderrepo", rootOpts.BuilderRepos, "list of docker repositories in descending priority order, used to search for builder images. Default falcosecurity/driverkit will always be enforced as lowest priority repo. eg: --builderrepo myorg/driverkit --builderrepo falcosecurity/driverkit")
+	flags.StringSliceVar(&rootOpts.BuilderRepos, "builderrepo", rootOpts.BuilderRepos, "list of docker repositories or file (absolute path) containing builder images index with the format '<image>,<target>,<gcc-version>[,<gcc-version>,...]', in descending priority order. Used to search for builder images. eg: --builderrepo myorg/driverkit --builderrepo falcosecurity/driverkit --builderrepo /path/to/my/index.txt. Index file line exmaple: yourorg/driverkit-builder;any;4.9.0;5.0.0;6.0.0;8.0.0")
```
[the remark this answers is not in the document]
FedeDP: Btw what if we used a structured text (like a yaml) instead of a txt file?
That way we could just parse it to a `[]Image` and be good to go!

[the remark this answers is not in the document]
FedeDP: (we can do this later before next release!)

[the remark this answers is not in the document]
Lowaiz: Hi, getting a yaml can be good yes, applying that right now !

FedeDP (approved):
/approve

PR #274, opened 2023-05-12
new: builder images tags via OCI

### LINKED ISSUE #256: Usage and break of builder image listing

FedeDP:
**What type of PR is this?**

/kind feature

/area cmd
/area pkg
/area docs

Multiple things are fixed:
* no more `docker search`; instead we will use falcoctl OCI utilities to retrieve list of tags for a docker repository
* moved our builder image name protocol to properly use a docker repo + tag (see below notes for reviewers)
* added back a default builder image tag, so that eg: driverkit releases will, by default, use builder images tagged together with them. This was a behavior once upon a time that was completely dropped some months ago. 
I think it still makes sense therefore i added it back.
* For `FileImageLister` make explicit the `arch` and `tag` shipped by an image entry
* Improved docs

Fixes #256 

CI will fail because we changed the builder image names protocol.
Old protocol:
* `falcosecurity/driverkit-builder-any-x86_64_gcc8.0.0_gcc6.0.0_gcc5.0.0_gcc4.9.0_gcc4.8.0:latest`

New protocol:
* `falcosecurity/driverkit-builder:any-x86_64_gcc8.0.0_gcc6.0.0_gcc5.0.0_gcc4.9.0_gcc4.8.0-latest`

As you can see, we now properly have a docker repository; to still be able to retrieve desired image tag (ie: latest) we had to add the `-latest` suffix.

As always, we might want to:
* either push the new builder images to a different repo (eg: `fededp/driverkit-builder`)
* or disable integration tests, merge this, and enable them back; this way master CI will take care of pushing new builder images for us

I'd go with the first (as it was the case other times we broke the builder images names), but i am currently using my smartphone internet data, therefore am not able to do so :) 

```release-note
new: fetch list of builder images via falcoctl OCI utilities
```


EXONER4TED:
The changes look good, I'm fine to just disable tests for this PR to let it build the new images for us.

EXONER4TED:
I'm not familiar with CircleCI... how do we disable those temporarily? Can you just comment them out in code? I suspect the project rules will require it though.

FedeDP:
We can just skip the `make integration_test` step: https://github.com/falcosecurity/driverkit/blob/master/.circleci/config.yml#LL19C7-L19C7

FedeDP:
I think we can wait a little bit more and then eventually consider any non-answer as yes :) 
Thank you!

FedeDP:
Dropped integration tests! :pray: for CI

FedeDP:
Imho we can go now @EXONER4TED ! :) Thank you!

FedeDP:
I added some more tests around `FileImagesLister` in latest commit!

EXONER4TED (approved):
🙏

PR #277, opened 2023-06-03
Fb update rhel9 example

toamto94:
**What type of PR is this?**

/kind documentation

/area docs

Only adds additional documentation which should make it easier to get started.

Added documentation for RHEL9 builder images where the underlying host is different. This is not covered in the documentation yet. Also, I added a line in the Dockerfile for RHEL9 images which fixes a bug that driverkit couldn't find the gcc-11 binary in rhel9 because a symlink from the redhat version of gcc-11 was missing.

No

```release-note
Added additional documentation for RHEL9 builder images
```

### DIFF
--- a/Example_configs.md
+++ b/Example_configs.md
@@ -269,9 +269,29 @@ and Dockerfile.rhel9:
 FROM docker.io/redhat/ubi9
 
 RUN yum install gcc elfutils-libelf-devel kmod make cpio llvm-toolset -y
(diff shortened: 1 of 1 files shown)


FedeDP:
Great catch, thank you very much!
Tagging @EXONER4TED to give this a look since he implemented rhel support :)

dwindsor (approved):

toamto94:
> Thanks for your contribution @toamto94! We were only able to get this to work on RHEL-based hosts, so this is very exciting.
> 
> Tagging @johnbrydon because he originally wrote this, but this LGTM.

Pleasure!

PR #279, opened 2023-06-05
fix(pkg): fixed mainline kernel URLs.

FedeDP:
**What type of PR is this?**

/kind bug

/area pkg

Moreover, properly enforce KBUILD_MODPOST_WARN in vanilla template.
It was previously enabled by default; but this commit: https://github.com/torvalds/linux/commit/5573b4daa26a0cf15aa0fecd7f1be16e0b6157bc merged in linux 6.3, changed the behavior.
Thus we now need to enforce it.

Fixes #

```release-note
NONE
```


dwindsor (approved):
Nice! Thanks

PR #285, opened 2023-08-28
cleanup(pkg): cleaned up URLs builder interface method removing config parameter

FedeDP:
**What type of PR is this?**

/kind cleanup

/area pkg

Reduce `URLs` builder interface method signature avoiding the unused `config` parameter.
It was only used by `ubuntu` builder to fetch build kernelversion; we can use viper for that.
This makes much more practical to use driverkit builder as library in other go projects, because you don't need to instantiate a full driverkit config to fetch URLs; for example:
```Go
var kernelheaders []string
builder, err := builder.Factory(builder.Type(myTarget))
if err != nil {
	logger.Warn(err.Error())
} else {
	kr := kernelrelease.FromString(myKernelRelease)
	kernelheaders, err = builder.URLs(kr)
	if err != nil {
		logger.Warn(err.Error())
	}
        logger.Info("Got headers!", kernelheaders)
}
```

Fixes #

```release-note
NONE
```


dwindsor:
Hey @FedeDP! This feature definitely makes sense - is there a companion MR for the viper-related changes for Ubuntu?

dwindsor:
> Hey @FedeDP! This feature definitely makes sense - is there a companion MR for the viper-related changes for Ubuntu?

Oh, nvm I see it =).

FedeDP:
I pushed a small commit to also expose `GetResolvingURLs` API :)

PR #289, opened 2023-09-04
chore(cmd,pkg,validate): drop logrus direct dependency.

FedeDP:
**What type of PR is this?**

/kind cleanup

/area cmd
/area pkg

Make use of golang 1.21 slog package.
Moreover, avoid enforcing a cobra.OnInitialize() function that would run even when driverkit is invoked as a library.

Fixes #

```release-note
NONE
```


.circleci/config.yml
```diff
@@ -2,14 +2,14 @@ version: 2.1
...
-      - image: alpine:3.18
+      - image: golang:1.21-alpine
     steps:
       - checkout
       - setup_remote_docker: # used by integration tests that runs driverkit binary that needs docker
           version: 20.10.24
       - run:
           name: Install deps
-          command: apk add gcc musl-dev make bash git go
+          command: apk add gcc musl-dev make bash git binutils-gold
```
FedeDP, line 12: Needed to fix a `ld` related issue on arm64

cmd/cli_test.go
```diff
@@ -302,6 +305,7 @@ func run(t *testing.T, test testCase) {
 	c := NewRootCmd()
 	b := bytes.NewBufferString("")
 	c.SetOutput(b)
+	validate.ProgramLevel.Set(slog.LevelDebug)
```
FedeDP, line 308: We want to enable debug level during tests.

cmd/config_options.go
```diff
@@ -28,7 +29,8 @@ type ConfigOptions struct {
 func NewConfigOptions() *ConfigOptions {
 	o := &ConfigOptions{}
 	if err := defaults.Set(o); err != nil {
-		logger.WithError(err).WithField("options", "ConfigOptions").Fatal("error setting driverkit options defaults")
+		slog.With("err", err.Error(), "options", "ConfigOptions").Error("error setting driverkit options defaults")
```
FedeDP, line 32: `slog` package has no "Fatal" level.
Emulate it through `os.Exit(1)` (that is exactly what Logrus.Fatal did).

dwindsor (approved):

PR #293, opened 2023-10-10
chore(cmd,pkg): skip build attempt when no artifacts are supported.

FedeDP:
**What type of PR is this?**

/kind cleanup

/area cmd
/area pkg

When we fixup builder output artifacts to just the supported ones, actually avoid doing a build when no artifact is supported.
This is rather similar to what is being done in `RootOptions.Validate`, BUT there we check that no artifact is supported at all, while in `ToBuild` we check whether the required artifact for the build is supported.
To quickly test the issue, just use this config:
```
kernelversion: 1
kernelrelease: 3.10.0-957.5.1.el7.x86_64
target: centos
architecture: amd64
output:
  probe: output/6.0.1+driver/x86_64/falco_centos_3.10.0-957.5.1.el7.x86_64_1.o
kernelurls: ["http://vault.centos.org/centos/7.6.1810/updates/x86_64/Packages/kernel-devel-3.10.0-957.5.1.el7.x86_64.rpm"]
```
Where a bpf probe only build is requested on an unsupported version; the `validate` step will pass because the kernelrelease supports the kmod, but then we have a no-artifact build going on that wastes time.

After this patch, this situation exits immediately with:
```
level=INFO msg="driver building, it will take a few seconds" processor=docker
level=WARN msg="Skipping build attempt of probe for unsupported kernel release" kernelrelease=3.10.0
```

Fixes #

```release-note
NONE
```


EXONER4TED (approved):
Seems reasonable to me!

PR #298, opened 2023-10-23
cleanup(cmd,pkg): properly manage ubuntu flavors at builders level.

FedeDP:
**What type of PR is this?**

/kind cleanup

/area cmd
/area pkg

This PR aims at resolving an issue we have while using driverkit as a go pkg: `ubuntu` flavors were translated in driverkit to just "ubuntu" in the `persistentValidateFunc` of root.go; this was the wrong place since while using driverkit as a pkg, you won't call those methods.
To be much cleaner, we now have:
* a write only, package local `builderByTarget` map, that has been renamed `builder.byTarget` (most of the lines change come from that)
* when we need read access to the map, we already had the `builder.Factory` method; it is now universally used, since the map is no more readable from outside. Here there is the logic to convert `ubuntu-$flavor` to `ubuntu` before getting the builder
* Side effect of the above: the log when building will now show real `target` that was requested (eg: `ubuntu-aws` instead of `ubuntu`)

Fixes #

```release-note
NONE
```


FedeDP:
Note, since this needs to be ported into https://github.com/falcosecurity/dbg-go to fix our drivers build pipeline, i will need to make a patch release for driverkit once this is merged :)

dwindsor (approved):
lgtm 👍

PR #302, opened 2023-11-10
wip: new(docker,pkg/driverbuilder): use `cmake` instead of makefile template to build kmod and bpf

FedeDP:
**What type of PR is this?**

/kind feature

/area pkg

Fixes #

```release-note
new(docker,pkg/driverbuilder): use cmake instead of makefile template to build kmod and bpf
```


docker/builders/builder-centos-x86_64_gcc4.8.5.Dockerfile
```diff
@@ -21,7 +21,15 @@ RUN yum -y install centos-release-scl && \
 	wget \
 	binutils \
 	which \
-	make
+	make \
+    git
+
+# Install cmake3.x (on centos7 `cmake` package installs cmake2.x)
+RUN curl -L -o /tmp/cmake.tar.gz https://github.com/Kitware/CMake/releases/download/v3.22.5/cmake-3.22.5-linux-$(uname -m).tar.gz; \
```
FedeDP, line 28: We need to manually install a cmake 3.x on centos:7.

pkg/driverbuilder/builder/templates/alinux.sh
```diff
@@ -28,10 +28,7 @@ rm -Rf /tmp/module-download
 mkdir -p /tmp/module-download
 
 curl --silent -SL {{ .ModuleDownloadURL }} | tar -xzf - -C /tmp/module-download
-mv /tmp/module-download/*/driver/* {{ .DriverBuildDir }}
-
-cp /driverkit/module-Makefile {{ .DriverBuildDir }}/Makefile
-bash /driverkit/fill-driver-config.sh {{ .DriverBuildDir }}
+mv /tmp/module-download/*/* {{ .DriverBuildDir }}
```
FedeDP, line 31: Scripts are a bit different now: 
* we move all the stuff under the download libs repo to`{{ .DriverBuildDir }}` now, not only the `driver` subfolder
* we run `cmake` with lots of things disabled. Unfortunately, given that libs only [recently](https://github.com/falcosecurity/libs/pull/1325) gained the ability to build the bpf probe in `MINIMAL_BUILD` mode, we need to configure the project in full mode, and this requires `git` because `grpc` and `protobuf` deps need to be fetched/cloned.
* Then, building is just a matter of calling `make driver` and `make bpf` with the correct `CC` and `KERNELDIR`.

FedeDP:
cc @falcosecurity/driverkit-maintainers

pkg/driverbuilder/builder/templates/alinux.sh
EXONER4TED: I actually really like how much this all simplifies the code, and I think the trade offs are worth it.

FedeDP:
I will also need to update the new `local`  builder to use `cmake` when downloading sources. (this is a reminder for myself!)

FedeDP:
Need more fixes to the new local (as in "internal to containers") module and probe paths (that will now be `/tmp/driver/build/driver/$drivername.ko` and `/tmp/driver/build/driver/bpf/probe.o`).
Everything else seems working fine though ;)

pkg/driverbuilder/builder/builders.go
```diff
@@ -55,6 +43,14 @@ type Config struct {
 	*Build
 }
 
+func (c Config) ToDriverFullPath() string {
+	return path.Join(DriverDirectory, "build", "driver", fmt.Sprintf("%s.ko", c.DriverName))
```
FedeDP, line 47: We expect the built artifact with cmake to be under `/tmp/driver/build/driver/$drivername.ko`.

pkg/driverbuilder/builder/templates/alinux.sh
```diff
@@ -42,19 +39,20 @@ rm -Rf /tmp/kernel
...
 
+cd {{ .DriverBuildDir }}
+mkdir -p build && cd build
+cmake -DUSE_BUNDLED_DEPS=On -DCREATE_TEST_TARGETS=Off -DBUILD_LIBSCAP_GVISOR=Off -DBUILD_LIBSCAP_MODERN_BPF=Off -DENABLE_DRIVERS_TESTS=Off -DDRIVER_NAME={{ .ModuleDriverName }} -DBUILD_BPF=On ..
+
 {{ if .BuildModule }}
 # Build the module
-cd {{ .DriverBuildDir }}
-make CC=/usr/bin/gcc-{{ .GCCVersion }} KERNELDIR=/tmp/kernel
-mv {{ .ModuleDriverName }}.ko {{ .ModuleFullPath }}
```
FedeDP, line 49: We don't need the `mv` anymore since ModuleFullPath is already the right path.

pkg/driverbuilder/kubernetes.go
```diff
@@ -102,11 +102,11 @@ func (bp *KubernetesBuildProcessor) buildModule(b *builder.Build) error {
 		return err
 	}
 
-	if builder.ModuleFullPath != "" {
+	if c.ModuleFilePath != "" {
```
FedeDP, line 105: The following are small fixes in the kubernetes builder. cc @Lowaiz

pkg/driverbuilder/kubernetes.go
```diff
@@ -275,15 +255,15 @@ func (bp *KubernetesBuildProcessor) copyModuleAndProbeFromPodWithUID(ctx context
 			}
 			if p.Status.Phase == corev1.PodRunning {
 				slog.With(falcoBuilderUIDLabel, falcoBuilderUID).Info("start downloading module and probe from pod")
-				if builder.ModuleFullPath != "" {
-					err = copySingleFileFromPod(build.ModuleFilePath, bp.coreV1Client, bp.clientConfig, p.Namespace, p.Name, builder.ModuleFullPath, moduleLockFile)
+				if c.ModuleFilePath != "" {
```
FedeDP, line 258: Same fix as above.

pkg/driverbuilder/kubernetes.go
```diff
@@ -275,15 +255,15 @@ func (bp *KubernetesBuildProcessor) copyModuleAndProbeFromPodWithUID(ctx context
...
+					err = copySingleFileFromPod(build.ModuleFilePath, bp.coreV1Client, bp.clientConfig, p.Namespace, p.Name, c.ToDriverFullPath(), moduleLockFile)
 					if err != nil {
 						return err
 					}
 					slog.Info("Kernel Module extraction successful")
 				}
-				if builder.ProbeFullPath != "" {
-					err = copySingleFileFromPod(build.ProbeFilePath, bp.coreV1Client, bp.clientConfig, p.Namespace, p.Name, builder.ProbeFullPath, probeLockFile)
+				if c.ProbeFilePath != "" {
```
FedeDP, line 265: Same fix as above.

FedeDP:
TODO: i need to also fix the `local` builder with local sources case.

FedeDP:
Done.

pkg/driverbuilder/builder/templates/alinux.sh
```diff
@@ -42,19 +39,20 @@ rm -Rf /tmp/kernel
 mkdir -p /tmp/kernel
 mv usr/src/kernels/*/* /tmp/kernel
 
+cd {{ .DriverBuildDir }}
+mkdir -p build && cd build
+{{ .CmakeCmd }}
```
FedeDP, line 44: Since we always use the same CmakeCmd, make it easily customizable (at least during development) by using a single templated variable for it.

pkg/driverbuilder/builder/builders.go
```diff
@@ -31,19 +31,25 @@ import (
...
-const ProbeFileName = "probe.o"
-
-// ModuleFullPath is the standard path for the kernel module. Builders must place the compiled module at this location.
-var ModuleFullPath = path.Join(DriverDirectory, ModuleFileName)
-
-// ProbeFullPath is the standard path for the eBPF probe. Builders must place the compiled probe at this location.
-var ProbeFullPath = path.Join(DriverDirectory, "bpf", ProbeFileName)
+const (
+	DriverDirectory = "/tmp/driver"
+	cmakeCmdFmt     = `cmake -Wno-dev \
```
FedeDP, line 36: This is the cmake command being run.
In the future we might want to allow users to override it perhaps? I don't think it will be needed anyway.

Btw it spots some unused CMake variable to keep compatibility with old driver versions, eg: `PROBE_NAME`, `PROBE_VERSION`...

pkg/driverbuilder/builder/builders.go
```diff
@@ -31,19 +31,25 @@ import (
...
+  -DDRIVER_NAME=%s \
+  -DPROBE_NAME=%s \
+  -DBUILD_BPF=On \
+  -DDRIVER_VERSION=%s \
+  -DPROBE_VERSION=%s \
+  -DGIT_COMMIT=%s \
+  -DDRIVER_DEVICE_NAME=%s \
+  -DPROBE_DEVICE_NAME=%s \
+  .. && \
+  sed -i s/'DRIVER_COMMIT ""'/'DRIVER_COMMIT "%s"'/g driver/src/driver_config.h`
```
FedeDP, line 51: In the end, we also manually set the `DRIVER_COMMIT` because currently libs do not honor `-DGIT_COMMIT` being set by external user: https://github.com/falcosecurity/libs/blob/master/cmake/modules/compute_versions.cmake#L35
Indeed, `GIT_COMMIT` is always overridden right now, therefore we force-set it in the driver config.

FedeDP: We could save all of this if we used `git` to download the repo and then checkout the needed version; but it would hurt download performances (ie: we would need to download the full repo history to allow `git` to compute versions for us, as we do in the libs CI: https://github.com/falcosecurity/libs/blob/master/.github/workflows/ci.yml#L61)

Makefile
```diff
@@ -108,7 +110,9 @@ integration_test: $(test_configs)
 
 .PHONY: $(test_configs)
 $(test_configs): ${driverkit}
-	${driverkit} docker -c $@ --builderimage auto:master -l debug --timeout 600
+	$(foreach d,$(DRIVERVERSIONS),\
```
FedeDP, line 113: We can now run integration tests against multiple driver versions.

FedeDP: This way, we can ensure backward compatibility for changes.

pkg/driverbuilder/builder/templates/sles.sh
FedeDP, line 1: @EXONER4TED i fixed up the sles builder too ;)

EXONER4TED:
So my vote here is to merge this, try it downstream, and then report back. I love this change, and we're ready to merge it. If something breaks downstream, we can handle it in a separate MR/release

EXONER4TED (approved):
/approve

PR #304, opened 2023-11-10
SUSE Linux Enterprise Support

EXONER4TED:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/area build

> /area cmd

> /area pkg

> /area docs

> /area tests

Adds initial support for SUSE Linux Enterprise. It's much like RHEL, will require a `--builderimage` with authentication setup, but it does allow the user the ability to bring-your-own-container and build.

Example:
```
scwx-sles12-sp5:~/driverkit # ~/driverkit/_output/bin/driverkit docker --kernelrelease 4.12.14-122.7.1 --kernelversion 1 --target sle --output-module ./test.ko --builderimage sle-builder:latest
level=INFO msg="driver building, it will take a few seconds" processor=docker
level=INFO msg="kernel module available" path=./test.ko
```

An example of a hacked-together Dockerfile used for the above build:
```
FROM registry.suse.com/suse/sles12sp5

COPY credentials/* /etc/zypp/credentials.d/

RUN zypper ref -s && zypper install -y SUSEConnect && SUSEConnect -p sle-module-toolchain/12/x86_64

RUN zypper ref && zypper --non-interactive install \
	make  \
	gcc   \
	tar   \
	gzip  \
	curl  \
	gcc8  \
	kmod && \
    ln -s /usr/bin/gcc-8 /usr/bin/gcc-8.0.0
```

Fixes # N/A

Crawling for SLES/D kernels should be relatively easy as well, using the same sort of strategy as RHEL with an authenticated container. I'll be making a PR to kernel-crawler as well.

```release-note
Adds SUSE Linux Enterprise support
```

### DIFF
--- a/pkg/driverbuilder/builder/sle.go
+++ b/pkg/driverbuilder/builder/sle.go
@@ -1,0 +1,64 @@
+// SPDX-License-Identifier: Apache-2.0
+/*
+Copyright (C) 2023 The Falco Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package builder
+
+import (
+	_ "embed"
+
+	"github.com/falcosecurity/driverkit/pkg/kernelrelease"
+)
+
+//go:embed templates/sle.sh
+var sleTemplate string
+
+// TargetTypeSLE identifies the sle target.
+const TargetTypeSLE Type = "sle"
+
+// sle is a driverkit target.
+type sle struct {
+}
+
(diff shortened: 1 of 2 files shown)


pkg/driverbuilder/builder/sle.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (v *sle) Name() string {
+	return TargetTypeSLE.String()
+}
+
+func (v *sle) TemplateScript() string {
+	return sleTemplate
+}
+
+func (v *sle) URLs(kr kernelrelease.KernelRelease) ([]string, error) {
```
FedeDP, line 50: Is there no way to fetch headers?

EXONER4TED: Not currently - SLE is an enterprise distro that requires authentication to get headers. For now, I followed what we did for RHEL, which is allow the user to bring-their-own-authenticated-container to build with `--builderimage`.

I'm working on an MR today for kernel-crawler to do the same as RHEL there too - allow the user to bring an auth'd container.

FedeDP: Oh i see, thanks! :D

FedeDP (approved):
/approve

PR #305, opened 2023-11-13
fix(pkg/driverbuilder): fixed docker container output when multiplexed.

FedeDP:
**What type of PR is this?**

/kind bug

/area pkg

When docker container attach is called on a non-tty terminal, stray characters will appear on screen because the docker SDK uses a custom multiplexing protocol allowing STDOUT and STDERR string to be sent to a single stream.
So, we now check if we are running multiplexed and adapt consequently.

Fixes #

```release-note
NONE
```


EXONER4TED (approved):
/approve

Issue #303, opened 2023-11-10
Failure in building kmod against recent archlinux kernels


FedeDP:
**Describe the bug**

Recent archlinux kmods fail to build (eg: against 6.6.1.arch1-1)
```
level=DEBUG msg="\x02\x00\x00\x00\x00\x00\x00\x7f./tools/objtool/objtool: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by ./tools/objtool/objtool)\n"
```
It seems we need a builder image with an updated glibc version.

PR #306, opened 2023-11-21
new(cmd,pkg): added new `local` build processor to build drivers using local kernel sources/gcc/clang

FedeDP:
**What type of PR is this?**

/kind feature

/area cmd
/area pkg

This PR introduces a new `local` build processor that will just build chosen driver version against local kernel sources using local gcc/clang.
Therefore, to work, it needs kernel-headers installed (eg: `/lib/modules/6.6.1-arch1-1/build`) and clang/gcc on the machine.
It does not use docker, of course.

Moreover, it supports building kmod through `dkms`, and supports directly using a provided preconfigured `src-dir`.
The bigger aim is to completely implement driver building logic that is today part of [`falco-driver-loader script`](https://github.com/falcosecurity/falco/blob/master/scripts/falco-driver-loader); this would allow us to properly use `driverkit` inside `falcoctl` to implement the `drivertype.Build` method (eg: https://github.com/falcosecurity/falcoctl/blob/main/pkg/driver/type/kmod.go#L183).
IMHO driverkit should be the only project responsible of building drivers.

Fixes #

```release-note
new: added a local build processor to build local kernel using local kernel sources/gcc/clang
```

### DIFF
--- a/cmd/local.go
+++ b/cmd/local.go
@@ -9,12 +9,20 @@ import (
 	"golang.org/x/sys/unix"
 	"log/slog"
 	"os"
+	"os/user"
 	"runtime"
 	"strings"
 )
 
+type localCmdOptions struct {
+	useDKMS bool
+	srcDir  string
+	envMap  map[string]string
+}
+
 // NewLocalCmd creates the `driverkit local` command.
 func NewLocalCmd(rootCommand *RootCmd, rootOpts *RootOptions, rootFlags *pflag.FlagSet) *cobra.Command {
+	opts := localCmdOptions{}
 	localCmd := &cobra.Command{
 		Use:               "local",
 		Short:             "Build Falco kernel modules and eBPF probes in local env with local kernel sources and gcc/clang.",
@@ -26,7 +34,18 @@ func NewLocalCmd(rootCommand *RootCmd, rootOpts *RootOptions, rootFlags *pflag.F
 				if !b.HasOutputs() {
 					return
 				}
-				if err := driverbuilder.NewLocalBuildProcessor(viper.GetInt("timeout")).Start(b); err != nil {
+				if opts.useDKMS {
+					currentUser, err := user.Current()
+					if err != nil {
+						slog.With("err", err.Error()).Error("Failed to retrieve user. Exiting.")
+						os.Exit(1)
+					}
+					if currentUser.Username != "root" {
+						slog.Error("Must be run as root for DKMS build.")
+						os.Exit(1)
+					}
+				}
+				if err := driverbuilder.NewLocalBuildProcessor(viper.GetInt("timeout"), opts.useDKMS, opts.srcDir, opts.envMap).Start(b); err != nil {
 					slog.With("err", err.Error()).Error("exiting")
 					os.Exit(1)
 				}
@@ -56,6 +75,9 @@ func NewLocalCmd(rootCommand *RootCmd, rootOpts *RootOptions, rootFlags *pflag.F
 			flagSet.AddFlag(flag)
 		}
 	})
+	flagSet.BoolVar(&opts.useDKMS, "dkms", false, "Enforce usage of DKMS to build the kernel module.")
+	flagSet.StringVar(&opts.srcDir, "src-dir", "", "Enforce usage of local source dir to build drivers.")
+	flagSet.StringToStringVar(&opts.envMap, "env", nil, "Env variables to be enforced during the driver build.")
 	localCmd.PersistentFlags().AddFlagSet(flagSet)
 	return localCmd
 }
--- a/cmd/local.go
+++ b/cmd/local.go
@@ -1,0 +1,84 @@
+package cmd
+
+import (
+	"bytes"
+	"github.com/falcosecurity/driverkit/pkg/driverbuilder"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+	"golang.org/x/sys/unix"
+	"log/slog"
+	"os"
+	"runtime"
+	"strings"
+)
+
+// NewLocalCmd creates the `driverkit local` command.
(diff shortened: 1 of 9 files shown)


cmd/local.go
```diff
@@ -0,0 +1,85 @@
...
+				}
+				if err := driverbuilder.NewLocalBuildProcessor(viper.GetInt("timeout")).Start(b); err != nil {
+					slog.With("err", err.Error()).Error("exiting")
+					os.Exit(1)
+				}
+			}
+		},
+	}
+	// Add root flags, but not the ones unneeded
+	unusedFlagsSet := map[string]struct{}{
```
FedeDP, line 37: These flags aren't needed for this command, therefore they are removed.
I think we might want to do the same eg: for `images` cmd.

cmd/local.go
```diff
@@ -0,0 +1,85 @@
...
+		if _, ok := unusedFlagsSet[flag.Name]; !ok {
+			flagSet.AddFlag(flag)
+		}
+	})
+	localCmd.PersistentFlags().AddFlagSet(flagSet)
+	return localCmd
+}
+
+// Partially overrides rootCmd.persistentPreRunFunc setting some defaults before config init/validation stage.
+func persistentPreRunFunc(rootCommand *RootCmd, rootOpts *RootOptions) func(c *cobra.Command, args []string) error {
```
FedeDP, line 64: This gets called in place of `rootCmd.PersistentPreRunE` since we shadowed it.
It does just set a bunch of default values (that are not used anyway by the `local` build processor, but are logged) and calls the `rootCmd.PersistentPreRunE` (ie: persistentValidateFunc)

pkg/driverbuilder/builder/builders.go
```diff
@@ -254,6 +255,13 @@ func Factory(target Type) (Builder, error) {
 	if strings.HasPrefix(target.String(), "ubuntu") {
 		target = Type("ubuntu")
 	}
+	// For fake "local" target (that is not exposed to users,
+	// nor registered in byTarget, just use "arch".
+	// NOTE: it won't be used anyway, as driverbuilder/local.go
+	// manually creates a "local" builder.
+	if target.String() == "local" {
```
FedeDP, line 262: This is a workaround to let `local` cmd set `local` target and passing the root options validation step.

Note: `local` builder is not mapped inside the `byTarget` map because i did not want it to be visible by the user when running eg: `docker` builder processor, since it makes no sense.
So, the new `local` builder is hidden and manually instantiated by `local` builder processor.

pkg/driverbuilder/builder/local.go
```diff
@@ -0,0 +1,53 @@
...
+
+// NOTE: since this is only used by local build,
+// it is not exposed in `target` array,
+// so no init() function to register it is present.
+
+//go:embed templates/local.sh
+var localTemplate string
+
+type LocalBuilder struct {
+	GccPath string
```
FedeDP, line 17: `GccPath` is the full path to a discovered Gcc, and will be set by `local` build processor.

FedeDP: Since there might be multiple `gcc` present, we tried all of them until we can build the module.

pkg/driverbuilder/builder/templates/local.sh
FedeDP, line 1: There is no kernel headers download step in the `local` script.

pkg/driverbuilder/local.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func (lbp *LocalBuildProcessor) Start(b *builder.Build) error {
+	slog.Debug("doing a new local build")
+
+	// We don't want to download headers
+	kr := b.KernelReleaseFromBuildConfig()
+
+	// create a builder based on the choosen build type
+	v := &builder.LocalBuilder{}
```
FedeDP, line 40: Manually instantiated; see https://github.com/falcosecurity/driverkit/pull/306/files#r1400657216

pkg/driverbuilder/local.go
```diff
@@ -0,0 +1,160 @@
...
+				continue
+			}
+			gccs = append(gccs, proposedGCC)
+		}
+	} else {
+		// We won't use it!
+		gccs = []string{"gcc"}
+	}
+
+	for _, gcc := range gccs {
```
FedeDP, line 102: For each discovered gcc, try to build module and probe.

cmd/local.go
```diff
@@ -0,0 +1,106 @@
...
+// NewLocalCmd creates the `driverkit local` command.
+func NewLocalCmd(rootCommand *RootCmd, rootOpts *RootOptions, rootFlags *pflag.FlagSet) *cobra.Command {
+	opts := localCmdOptions{}
+	localCmd := &cobra.Command{
+		Use:               "local",
+		Short:             "Build Falco kernel modules and eBPF probes in local env with local kernel sources and gcc/clang.",
+		PersistentPreRunE: persistentPreRunFunc(rootCommand, rootOpts),
+		Run: func(c *cobra.Command, args []string) {
+			slog.With("processor", c.Name()).Info("driver building, it will take a few seconds")
+			if !configOptions.DryRun {
```
dwindsor, line 32: Just a stylistic nit: you could remove a level of indentation in this closure with 

```go
if configOptions.DryRun {
    return
}
```

dwindsor:
LGTM! Incorporating `driverkit` into `falcoctl` is a great idea.

Is `falcoctl` the only intended user of the `local` build processor? IOW, will we ever expect an end user to actually compile w/ local headers?

dwindsor (approved):

FedeDP:
> Is falcoctl the only intended user of the local build processor? IOW, will we ever expect an end user to actually compile w/ local headers?

Well, since eg: Arch builder is hit by https://github.com/falcosecurity/driverkit/issues/303, i was able to build the drivers on my machine using the new local build processor :D 
But yep, i think the user will be mostly falcoctl!

PR #307, opened 2023-11-30
cleanup(cmd): require kernelrelease and kernelversion in local command.

FedeDP:
**What type of PR is this?**

/kind cleanup

/area cmd

Require `kernelrelease` and eventually `kernelversion` to be passed by user instead of using `unix`. This allows driverkit to be portable (eg: building on windows).
Useful to fix https://github.com/falcosecurity/falcoctl/pull/356 since falcoctl is built for windows and osx too and its deps must be portable.

It is a small UX worsening but since all driverkit commands already require `kernelrelease`, i think it is safe to do so for `local` too.

Fixes #

```release-note
NONE
```

### DIFF
--- a/cmd/local.go
+++ b/cmd/local.go
@@ -1,17 +1,14 @@
 package cmd
 
 import (
-	"bytes"
 	"github.com/falcosecurity/driverkit/pkg/driverbuilder"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
-	"golang.org/x/sys/unix"
 	"log/slog"
 	"os"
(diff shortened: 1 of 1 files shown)


FedeDP:
Most probably, as next step for the `local` builder, i would love to introduce back the capability to receive `--kernelurls` parameter and download them.
Here the hardest part would be to correctly extract and manage them since kernel headers packages can have different format on different distros and we have a single `local` builder template.
I think that the `local` builder processor can download kernelurls packages for us; then we might want to use eg: https://github.com/mholt/archiver to extract the kernel headers packages, no matters the format.

PR #308, opened 2023-12-01
Tweaking SLES Builder

EXONER4TED:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area cmd

/area pkg

> /area docs

> /area tests

Tweaking the SLES builder to:

1. match the OS ID (which is `sles` actually, when I thought it was just `sle` previously):
```
scwx-sles12-sp5:/ # cat /etc/os-release
NAME="SLES"
VERSION="12-SP5"
VERSION_ID="12.5"
PRETTY_NAME="SUSE Linux Enterprise Server 12 SP5"
ID="sles"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:12:sp5"
```
2. The SLES builder requires `--net=host` on the docker CLI at runtime to properly connect the container to the host for authentication. It's kind of odd, but more info can be found on it here: https://github.com/SUSE/container-suseconnect 

For using "pay as you go" images in public clouds for SUSE, the OS will authenticate to SUSE mirrors automatically. Any container on the host will see this authentication as well automatically with the `containerbuild-regionsrv` service. However, for any of that to work, the container _must_ run in host network mode.

Fixes # N/A

I'm only halfway certain this will work, so more PR's may be needed after this one if it doesn't work. Currently in a bit of a trial/error on the authentication for SLES, but this change should be harmless since I'm only catching the SLES target type (which doesn't really work anyway currently).

Nope

```release-note
NONE
```


pkg/driverbuilder/docker.go
```diff
@@ -203,9 +205,17 @@ func (bp *DockerBuildProcessor) Start(b *builder.Build) error {
 		Image: builderImage,
 	}
 
+	// check if on a sles target type, which requires docker to run with `--net=host` for builder images to work
+	// for more info, see the suse container connect README: https://github.com/SUSE/container-suseconnect
+	var netMode = container.NetworkMode("default")
+	if b.TargetType == "sles" {
```
FedeDP, line 211: I would add a new builders interface (Network mode() string) and only implement in for sles builder instead.
Here we can just check
```
if vv, ok := v.(NetworkModder); ok {
  NetMode = vv.NetworkMode()
}
```
I think it is more elegant and future proof this way because of the builder knowledge will belong to the builder specific source file.

EXONER4TED: This makes way more sense 🙂 I just removed the if and set the default network mode to `"default"`, but it can be overridden in an individual builder. Then the hostConfig just has:
```
NetworkMode: container.NetworkMode(b.BuilderImageNetworkMode),
```

Let me know if you'd like it a different way or want the functions moved around. I put the default function in `builders.go`

FedeDP (approved):
/approve

PR #310, opened 2023-12-12
fix(pkg): fixed docker multiplexed output function.

FedeDP:
**What type of PR is this?**

/kind bug

/area pkg

The fixed multiplexed docker output merged in #305 had some bugs when the `Read` for the output was truncated, since the case was not managed. 
This is breaking libs CI since the driverkit v0.16.0 tag: https://github.com/falcosecurity/libs/actions/workflows/latest-kernel.yml

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/driverbuilder/docker.go
+++ b/pkg/driverbuilder/docker.go
@@ -29,6 +29,7 @@ import (
 	"os"
 	"runtime"
 	"strconv"
+	"strings"
 
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
@@ -286,7 +287,7 @@ func (bp *DockerBuildProcessor) Start(b *builder.Build) error {
 		return err
 	}
 
-	hr, err := cli.ContainerExecAttach(ctx, edata.ID, types.ExecStartCheck{Tty: false})
+	hr, err := cli.ContainerExecAttach(ctx, edata.ID, types.ExecStartCheck{})
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


pkg/driverbuilder/docker.go
```diff
@@ -396,20 +397,43 @@ func forwardLogs(logPipe io.Reader) {
...
 	hdr := make([]byte, 8)
 	for {
+		// Load size of message
 		_, err := logPipe.Read(hdr)
 		if err == io.EOF {
 			break
 		}
 		if err != nil {
 			slog.With("err", err.Error()).Error("log pipe error")
+			return
```
FedeDP, line 407: Return in case of errors.

pkg/driverbuilder/docker.go
```diff
@@ -396,20 +397,43 @@ func forwardLogs(logPipe io.Reader) {
...
 		}
 		count := binary.BigEndian.Uint32(hdr[4:])
+
+		// Read message
 		dat := make([]byte, count)
-		_, err = logPipe.Read(dat)
-		if err != nil {
-			slog.With("err", err.Error()).Error("log pipe error")
+		var readCnt int
+		for uint32(readCnt) < count {
```
FedeDP, line 414: Read the message in multiple `Read` until we read all bytes.
For big messages (ie >4096) , they might get split and we need to manage it.

pkg/driverbuilder/docker.go
```diff
@@ -396,20 +397,43 @@ func forwardLogs(logPipe io.Reader) {
...
+				return
+			}
+			if err != nil {
+				slog.With("err", err.Error()).Error("log pipe error")
+				return
+			}
+		}
+
+		// Print message line by line
+		lines := strings.Split(string(dat), "\n")
```
FedeDP, line 431: Instead of printing all the output in a single log, split it line by line and print it.

PR #319, opened 2024-02-14
fix(pkg/driverbuilder): fixed local cmd build up using os.Environ().

FedeDP:
**What type of PR is this?**

/kind bug

/area pkg

Moreover, properly mount debugfs before attempting the bpf probe build if needed.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/driverbuilder/builder/templates/local.sh
+++ b/pkg/driverbuilder/builder/templates/local.sh
@@ -59,6 +59,12 @@ modinfo {{ .ModuleFullPath }}
 
 {{ if .BuildProbe }}
 echo "* Building eBPF probe"
+
+if [ ! -d /sys/kernel/debug/tracing ]; then
+  echo "* Mounting debugfs"
+  mount -t debugfs nodev /sys/kernel/debug
+fi
+
 # Build the eBPF probe
 cd {{ .DriverBuildDir }}/bpf
 make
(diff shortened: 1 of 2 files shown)


pkg/driverbuilder/builder/templates/local.sh
```diff
@@ -59,6 +59,12 @@ modinfo {{ .ModuleFullPath }}
 
 {{ if .BuildProbe }}
 echo "* Building eBPF probe"
+
+if [ ! -d /sys/kernel/debug/tracing ]; then
```
FedeDP, line 63: If needed, mount debugfs before locally building the probe.

pkg/driverbuilder/local.go
```diff
@@ -128,6 +128,7 @@ func (lbp *LocalBuildProcessor) Start(b *builder.Build) error {
 		ctx, cancelFunc := context.WithTimeout(context.Background(), time.Duration(lbp.timeout)*time.Second)
 		defer cancelFunc()
 		cmd := exec.CommandContext(ctx, "/bin/bash", "-c", driverkitScript)
+		cmd.Env = os.Environ()
```
FedeDP, line 131: If `cmd.Env` is not empty when running the cmd, it won't get auto-filled with `Environ`. Force fill before appending our env map.

EXONER4TED (approved):
/approve

PR #321, opened 2024-02-29
wip: chore(test): added a centos_3.x regression test to highlight the bug.

FedeDP:
**What type of PR is this?**

/kind bug

/area build
/area tests

Fixes #

```release-note
NONE
```


FedeDP:
For now, this just highlights the bug by adding a new regression test config specifically for centos 3.x that uses the `driverkit-builder:centos-x86_64_gcc4.8.5-$tag` image.

FedeDP:
Most probably, when run by driverkit, the `scl_source enable llvm-toolset-7.0` line (https://github.com/falcosecurity/driverkit/blob/master/docker/builders/builder-centos-x86_64_gcc4.8.5.Dockerfile#L37C11-L37C46) is not correctly parsed from `/etc/bashrc` therefore we end up with a broken environment.

pkg/driverbuilder/docker.go
```diff
@@ -259,6 +259,7 @@ func (bp *DockerBuildProcessor) Start(b *builder.Build) error {
 		Env:          envs,
 		Cmd: []string{
 			"/bin/bash",
+			"-l",
```
FedeDP, line 262: By setting the "login" flag to bash, we enforce it to load data from bashrc/bash profiles.
See https://stackoverflow.com/questions/55206227/why-bashrc-is-not-executed-when-run-docker-container.

EXONER4TED (approved):

PR #333, opened 2024-04-11
new(cmd,pkg,docs,docgen,validate): switched from `slog` go library to rich-text falcoctl log library

FedeDP:
**What type of PR is this?**

/kind feature

/area cmd
/area pkg
/area docs

To better integrate falcoctl and driverkit, we needed to:
* avoid using a global logger in driverkit (we always used default `slog` instance)
* possibly use same logger in both library

Since falcoctl log library is much richer and better, i ported driverkit to use it; now each executor takes a logger parameter, that makes much easier and seamless the ingreation with falcoctl.
The result is wonderful:
[driverkit.webm](https://github.com/falcosecurity/driverkit/assets/5837210/dc4030a0-06f4-47b0-b659-a30eb615ab42)

While i was at it, i also:
* updated docs (never updated since the introduction of `local` builder
* ported driverkit `help/usage` message printing to follow same conventions as falcoctl (ie: print them less often basically :D)

TODO:
- [x] fix tests :)

Fixes #

```release-note
new(cmd,pkg,docs,docgen,validate): switched from `slog` go library to rich-text falcoctl log library
```


cmd/completion.go
```diff
@@ -46,12 +47,12 @@ func validateArgs() cobra.PositionalArgs {
 		if len(args) == 0 {
 			return nil
 		}
-		return cobra.ExactValidArgs(1)(c, args)
+		return cobra.MatchAll(cobra.ExactArgs(1), cobra.OnlyValidArgs)(c, args)
```
FedeDP, line 50: This avoids using a deprecated method.

cmd/completion.go
```diff
@@ -46,12 +47,12 @@ func validateArgs() cobra.PositionalArgs {
...
 	}
 }
 
 // NewCompletionCmd ...
-func NewCompletionCmd() *cobra.Command {
+func NewCompletionCmd(_ *ConfigOptions, _ *RootOptions, _ *pflag.FlagSet) *cobra.Command {
```
FedeDP, line 55: Same signature as other commands.

cmd/config_options.go
```diff
@@ -15,51 +15,123 @@ limitations under the License.
...
 type ConfigOptions struct {
 	ConfigFile string
-	LogLevel   string `validate:"loglevel" name:"log level" default:"INFO"`
 	Timeout    int    `validate:"number,min=30" default:"120" name:"timeout"`
 	ProxyURL   string `validate:"omitempty,proxy" name:"proxy url"`
 	DryRun     bool
 
-	configErrors bool
+	// Printer used by all commands to output messages.
+	Printer *output.Printer
```
FedeDP, line 46: TODO: double check if `Printer` and `Writer` can be package private.

cmd/config_options.go
```diff
@@ -15,51 +15,123 @@ limitations under the License.
...
 
 // Validate validates the ConfigOptions fields.
-func (co *ConfigOptions) Validate() []error {
+func (co *ConfigOptions) validate() []error {
 	if err := validate.V.Struct(co); err != nil {
-		errors := err.(validator.ValidationErrors)
-		errArr := []error{}
-		for _, e := range errors {
+		var errs validator.ValidationErrors
+		errors.As(err, &errs)
```
FedeDP, line 75: Use `errors.As` as suggested by ide.
This kind of cleanup is done multiple times.

pkg/driverbuilder/builder/amazonlinux.go
```diff
@@ -212,7 +211,6 @@ func buildMirror(a amazonBuilder, r string, kv kernelrelease.KernelRelease) (str
 	}
 
 	mirror := fmt.Sprintf("%s/%s", baseURL, "mirror.list")
-	slog.With("url", mirror, "version", r).Debug("looking for repo...")
```
FedeDP, line 215: I don't think we actually needed these ones; passing a `printer` down there would be a nightmare :) 
Also, i think distro builders should not log anything.

pkg/driverbuilder/builder/builders.go
```diff
@@ -415,7 +421,6 @@ func GetResolvingURLs(urls []string) ([]string, error) {
 		}
 		if res.StatusCode == http.StatusOK {
 			results = append(results, u)
-			slog.With("url", u).Debug("kernel header url found")
```
FedeDP, line 418: Again, passing the printer down here would be a bit troublesome.
I preferred to just kill this debug log (and adding a log in `KernelDownloadScript`: https://github.com/falcosecurity/driverkit/pull/333/files#diff-16a66a3bc20a8ec693f64ed821148f71efa386a7e942444444eb591512bf129bR173)

pkg/driverbuilder/builder/builders.go
```diff
@@ -389,13 +397,11 @@ func (c Config) toTemplateData(b Builder, kr kernelrelease.KernelRelease) common
...
 	if err != nil {
-		slog.Error(err.Error())
-		os.Exit(1)
+		panic(err)
 	}
 	base, err := url.Parse(uu.Host)
 	if err != nil {
-		slog.Error(err.Error())
-		os.Exit(1)
+		panic(err)
```
FedeDP, line 404: Panic instead of erroring out. Again, this was done to avoid passing the printer down here.

cmd/docker.go
```diff
@@ -15,32 +15,37 @@ limitations under the License.
...
-func NewDockerCmd(rootOpts *RootOptions, rootFlags *pflag.FlagSet) *cobra.Command {
+func NewDockerCmd(configOpts *ConfigOptions, rootOpts *RootOptions, rootFlags *pflag.FlagSet) *cobra.Command {
 	dockerCmd := &cobra.Command{
 		Use:   "docker",
 		Short: "Build Falco kernel modules and eBPF probes against a docker daemon.",
-		Run: func(c *cobra.Command, args []string) {
-			slog.With("processor", c.Name()).Info("driver building, it will take a few seconds")
-			if !configOptions.DryRun {
-				b := rootOpts.ToBuild()
+		RunE: func(c *cobra.Command, args []string) error {
```
FedeDP, line 29: Another small change: return an error from executors instead of dealing with `os.Exit(1)` ourselves.

cmd/docker.go
```diff
@@ -15,32 +15,37 @@ limitations under the License.
...
+			configOpts.Printer.Logger.Info("starting build",
+				configOpts.Printer.Logger.Args("processor", c.Name()))
+			if !configOpts.DryRun {
+				// Since we use a spinner, cache log data to a bytesbuffer;
+				// we will later print it once we stop the spinner.
+				var buf bytes.Buffer
+				b := rootOpts.ToBuild(configOpts.Printer.WithWriter(&buf))
```
FedeDP, line 36: This is a small trick: since when using a `Spinner` we **cannot** print anything to screen (otherwise that would cause a spinner refresh that changes its title and looks ugly), we store to `Build` a printer that prints to a bytes buffer; then we defer a function that prints everything **after** the spinner was stopped.

cmd/config_options.go
EXONER4TED: These changes all make sense to me, but commenting here just to make sure you've resolved this TODO (if you wanted to in this MR)

FedeDP: Yes! I could not make the change :)

FedeDP:
Needs a rebase! Going to rebase asap!

PR #394, opened 2025-07-29
chore(deps): migrate blang/semver to v4

dnwe:
**What type of PR is this?**

/kind cleanup

/area pkg

The semver package moved to the versioned import path some time back. Move to /v4 to sync with falcoctl and avoid pulling in two copies of the module.

```release-note
NONE
```


FedeDP (approved):
Thanks for the cleanup!
/approve

